PR #1, opened 2017-07-02
Update README.md

slang25:
As someone who's had to deal with this exact problem before this library existed, let me say thank you for this library!

I've added code syntax highlighting to the readme.


RehanSaeed:
I tweeted a minute ago. Fastest PR ever!

PR #4, opened 2017-11-16
Support deserialization

psampaio:
This PR is an initial implementation of the deserialization of Schema.org object, by adding the ReadJson method to both converters. I'll try to make several comments in the files below, but here's the general overview:

- It deserializes Values\<T\> classes
- It deserializes the remaining Values<T...> classes, but there's a fallback to figure out the right T for the type of persisted data
- It deserializes Enums
- It respects the @type properties, when they exist
- It supports generic lists and figures out the correct derived type, i.e., List\<CreativeWork\> with Book elements
- It has unit tests, although they rely on ToString too much. This is because it's not trivial (and maybe doesn't even make sense) to implement Equals
- I've kept the unit tests in a separate commit, because I did some structural changes that can be bit aggressive to the original code
- It relies on Activator and GetTypeInfo to get and instantiate the correct types, so the performance may need tweaking

Hope it helps

### DIFF
--- a/Source/Schema.NET/TimeSpanToISO8601DurationValuesConverter.cs
+++ b/Source/Schema.NET/TimeSpanToISO8601DurationValuesConverter.cs
@@ -1,6 +1,7 @@
-﻿namespace Schema.NET
+namespace Schema.NET
 {
     using System;
+    using System.Reflection;
     using System.Xml;
     using Newtonsoft.Json;
 
@@ -11,6 +12,34 @@
     /// <seealso cref="JsonConverter" />
     public class TimeSpanToISO8601DurationValuesConverter : ValuesConverter
     {
+        /// <inheritdoc />
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var valuesType = objectType.GetTypeInfo().IsGenericType && objectType.GetGenericTypeDefinition() == typeof(Nullable<>)
+                ? Nullable.GetUnderlyingType(objectType)
+                : objectType;
+
+            if (valuesType.GenericTypeArguments.Length == 1)
+            {
+                var mainType = valuesType.GenericTypeArguments[0];
+                Type genericType = typeof(Values<TimeSpan>);
+                if (mainType.GetTypeInfo().IsGenericType && mainType.GetGenericTypeDefinition() == typeof(Nullable<>))
+                {
+                    mainType = Nullable.GetUnderlyingType(mainType);
+                    genericType = typeof(Values<TimeSpan?>);
+                }
+
+                if (mainType == typeof(TimeSpan))
+                {
+                    var timeSpan = XmlConvert.ToTimeSpan(reader.Value.ToString());
+                    var instance = Activator.CreateInstance(genericType, timeSpan);
+                    return instance;
+                }
+            }
+
+            return base.ReadJson(reader, objectType, existingValue, serializer);
+        }
+
         /// <summary>
         /// Writes the object retrieved from <see cref="IValue"/> when one is found.
         /// </summary>
--- a/Source/Schema.NET/TimeSpanToISO8601DurationValuesConverter.cs
+++ b/Source/Schema.NET/TimeSpanToISO8601DurationValuesConverter.cs
@@ -15,15 +15,13 @@ namespace Schema.NET
         /// <inheritdoc />
         public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
         {
-            var valuesType = objectType.GetTypeInfo().IsGenericType && objectType.GetGenericTypeDefinition() == typeof(Nullable<>)
-                ? Nullable.GetUnderlyingType(objectType)
-                : objectType;
+            var valuesType = objectType.GetUnderlyingTypeFromNullable();
 
             if (valuesType.GenericTypeArguments.Length == 1)
             {
                 var mainType = valuesType.GenericTypeArguments[0];
-                Type genericType = typeof(Values<TimeSpan>);
-                if (mainType.GetTypeInfo().IsGenericType && mainType.GetGenericTypeDefinition() == typeof(Nullable<>))
+                var genericType = typeof(Values<TimeSpan>);
+                if (mainType.IsNullable())
                 {
                     mainType = Nullable.GetUnderlyingType(mainType);
                     genericType = typeof(Values<TimeSpan?>);
--- a/Source/Schema.NET/ValuesConverter.cs
+++ b/Source/Schema.NET/ValuesConverter.cs
@@ -147,7 +147,7 @@ namespace Schema.NET
                             }
                             else
                             {
-                                args = Activator.CreateInstance(type, value);
+                                args = token.ToObject(type); // This is expected to throw on some case
                             }
                         }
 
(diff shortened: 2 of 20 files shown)


Source/Schema.NET/TimeSpanToISO8601DurationValuesConverter.cs
```diff
@@ -11,6 +12,34 @@
     /// <seealso cref="JsonConverter" />
     public class TimeSpanToISO8601DurationValuesConverter : ValuesConverter
     {
+        /// <inheritdoc />
```
psampaio: This unwraps the nullable, deserializes the TimeSpan with the same XmlConverter used to serialize and wraps again in a Values<TimeSpan> or its nullable version

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+        {
+            var mainType = UnwrapNullable(objectType);
+
+            object argument = null;
+
+            var tokenType = reader.TokenType;
+            var value = SanitizeReaderValue(reader, tokenType);
+
+            JToken token = JToken.Load(reader);
+            if (mainType.GenericTypeArguments.Length == 1)
```
psampaio: There are two major blocks: Values<T> and the other Values<T....>. For this one we only need to figure the correct type and set "argument" accordingly, to be used in the end return.

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+            {
+                var type = UnwrapNullable(mainType.GenericTypeArguments[0]);
+                if (tokenType == JsonToken.StartArray)
+                {
+                    var listType = typeof(List<>).MakeGenericType(type);
+                    if (token.Any(t => !string.IsNullOrEmpty(GetTypeNameFromToken(t))))
```
psampaio: If the collection has a child with the @type attribute, we create a list with the base type, create each child with the correct type and add it. We need to resort to reflection, because the most common base class would be Thing and it clashes with the list type.

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                    else
+                    {
+                        argument = token.ToObject(listType);
+                    }
+                }
+                else if (IsPrimitiveType(type))
+                {
+                    argument = value;
+                }
+                else if (type == typeof(decimal))
```
psampaio: There are a couple of cases like this: there's not decimal on JSON without outputting more type information. But since we know that the target type is decimal, we convert it.

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                {
+                    argument = Convert.ToDecimal(value);
+                }
+                else
+                {
+                    if (type.GetTypeInfo().IsEnum)
```
psampaio: Parse the Enums from "http://schema.org/{EnumValue}"

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                    {
+                        var enumString = token.ToString().Substring("http://schema.org/".Length);
+                        argument = Enum.Parse(type, enumString);
+                    }
+                    else
+                    {
+                        var typeName = GetTypeNameFromToken(token);
```
psampaio: Again, if we have the @type attribute, use that instead

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                        {
+                            var builtType = Type.GetType($"Schema.NET.{typeName}");
+                            argument = token.ToObject(builtType);
+                        }
+                    }
+                }
+            }
+            else
+            {
+                foreach (var type in mainType.GenericTypeArguments)
```
psampaio: This is the block that handles the other Values<T....>. Here we need to iterate all the allowed types to see which one is a match.

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                            {
+                                args = token.ToObject(type);
+                            }
+                            else if (typeName == type.Name)
+                            {
+                                args = token.ToObject(type);
+                            }
+                            else
+                            {
+                                var builtType = Type.GetType($"Schema.NET.{typeName}");
```
psampaio: This handles the case where the T is a base class, like when T is Place but the actual type is Country

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                            {
+                                args = Convert.ToDecimal(value);
+                            }
+                            else
+                            {
+                                args = token.ToObject(type); // This is expected to throw on some case
+                            }
+                        }
+
+                        var genericType = typeof(Values<>).MakeGenericType(type);
```
psampaio: Here we create the correct Values<T> to set

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -40,8 +34,147 @@ public class ValuesConverter : JsonConverter
...
+                            {
+                                args = value;
+                            }
+                            else if (unwrappedType == typeof(decimal))
+                            {
+                                args = Convert.ToDecimal(value);
+                            }
+                            else
+                            {
+                                args = token.ToObject(type); // This is expected to throw on some case
```
psampaio: This is a brute force fallback for the cases when we have a primitive value to deserialize, but the types are all complex, i.e., we have a string with an URL, but the types are ImageObject or Uri. In this case, string is a valid parameter for an Uri constructor so that works. This will throw when trying to create an ImageObject with a string parameter, which will be caught below

Source/Schema.NET/ValuesConverter.cs
```diff
@@ -74,5 +207,28 @@ public virtual void WriteObject(JsonWriter writer, object value, JsonSerializer
...
+        }
+
+        private static bool IsPrimitiveType(Type type)
+        {
+            return type.GetTypeInfo().IsPrimitive || type == typeof(string);
+        }
+
+        private static object SanitizeReaderValue(JsonReader reader, JsonToken tokenType)
+        {
+            return tokenType == JsonToken.Integer ? Convert.ToInt32(reader.Value) : reader.Value;
```
psampaio: Integers will always come as Int64. We probably need a more resilient way to handle, but have not figured it out yet.

RehanSaeed:
Wow, this is a great PR to have. I'll need to find some time to digest it to see if we can get it tweaked and/or merged.

RehanSaeed:
There is one more try catch than I would have liked but it passes the tests and the basic usefulness is there. Good job! The package will be on MyGet in a few minutes (link on the front page), give it a test and I'll push it out soon.

psampaio:
Thanks for accepting the PR. I'll test it soon (probably over the weekend, only) and provide feedback if needed.

RehanSaeed:
Excellent!

Issue #5, opened 2018-02-16
As far as I can see not possible to mix types in Values{T1,....Tn}


AndreSteenbergen:
I am trying to make a parser from html microdata to Schema.Net. As far as I can see now, I can only use 1 type in Values with more then 1 generic argument.

e.g. I would like to be able to parse an image using url, but somewhere later in the document I would like to add ImageObject type, because the creator of the html page made a differnce in types of images.

AndreSteenbergen:
Must say: nice class generating tool. I changed two classes to suit my needs better. I think it should be possible now to mix Values, and so loosing the Values{T1,..Tn} objects. I wouldn't know a good way to do newtonsoft serialization. 

I added the following method to property, might be useful to someone


        public void AppendIndentLineInterface (StringBuilder stringBuilder, int indent)
        {
            stringBuilder.AppendIndentLine(indent, $"public interface I{this.Name} {{ object Data {{ get; set; }} }}");
            stringBuilder.AppendIndentLine(indent, $"public interface I{this.Name}<T> : I{this.Name} {{ new T Data {{ get; set; }} }}");
            foreach (var type in this.Types)
            {
                stringBuilder.AppendIndentLine(indent, $"public class {this.Name}{type.CSharpTypeString} : I{this.Name}<{type.CSharpTypeString}>");
                stringBuilder.AppendIndentLine(indent, $"{{");
                stringBuilder.AppendIndentLine(indent + 4, $"object I{this.Name}.Data => Data;");
                stringBuilder.AppendIndentLine(indent + 4, $"public virtual {type.CSharpTypeString} Data {{ get; set; }}");
                stringBuilder.AppendIndentLine(indent + 4, $"public {this.Name}{type.CSharpTypeString} () {{ }}");
                stringBuilder.AppendIndentLine(indent + 4, $"public {this.Name}{type.CSharpTypeString} ({type.CSharpTypeString} data) {{ Data = value; }}");
                stringBuilder.AppendIndentLine(indent + 4, $"public static implicit operator {this.Name}{type.CSharpTypeString} ({type.CSharpTypeString} data) {{ return new {this.Name}{type.CSharpTypeString} (data); }}");
                stringBuilder.AppendIndentLine(indent, $"}}");
                stringBuilder.AppendLine();
            }
        }
```

AndreSteenbergen:
You end with classes like this, the Values<T> can now be of type IIdentifer, I made the interfaces and classes subclasses to the type.

[...]

Issue #2, opened 2017-08-05
Using Schema.NET in ASP.NET Core controllers in model binding


veikkoeeva:
Hi!

I was wondering if you would have advice on how to use this interesting library in ASP.NET Core controllers so that roughly the following would work:

```
[HttpPost("[action]")]
public IActionResult Test([FromBody] Thing thing)
{
    var e = (Person)thing;
    return Accepted(new Uri("https://wwww.WhereToAskForResults.com"));
}
```
That is, accept JSON-LD data via base classes.

RehanSaeed:
I had a [similar conversation](https://twitter.com/MattHidinger/status/892094057391857664) with someone over Twitter a couple of days ago. The library only supports serialization from objects to JSON-LD at the moment and not deserialization from JSON-LD to object.

What's your use case? I think it should be fairly simple to add deserialization support. The Tweeter said he would try to add support and maybe submit a PR.

veikkoeeva:
@RehanSaeed That would be a very welcomed addition. My use case is actually one you see in the code snippet, a web service that ingests JSON-LD data of various sorts, but thinking further, there are at least two parts to it:

1. Have a library that can read plain-text JSON-LD documents and transform them to strongly typed classes.
2. Have middleware or appropriate attributes to place that to ASP.NET Core pipeline.

I personally have the need only for the _compacted form_.

As the format is self-describing, it makes sense to be able to ingest it via base classes and then do pattern matching to the actual type. This brings one design consideration: Maybe it makes sense that if one doesn't have the most derived types present, some less derived class in the hierarchy could be used while the "excess" would be available somewhow for manipulation. The reason for this is that JSON-LD is a self-describing format so one could expect to receive all kinds of data and maybe not (yet) have the concrete, most derived types loaded into process at that process and/or point in time. Though this is more like a feature consideration now, maybe useful for discussion.

Taking a peek at the Twitter conversation, I took a look at the code too. I had this exactly same idea that the `@type` is the same as class together with the context being more like a namespacey thingy and those could be used as a discriminator. It would be very useful if the tooling you've developed already could be opened and adapted to ingest other than _schema.org_ data to generate strong types (and make them have `IEquatable`, `IStructuralEquatable` etc. if possible).

Spelunking a bit through Internet, it looks there are efforts to have this functionality already, I don't know how useable they are. Some examples:
- https://github.com/wikibus/JsonLD.Entities
- https://github.com/NuGet/json-ld.net/issues/3
[...]

PR #6, opened 2018-02-19
Changed the way multitype values work, and included a html parser

AndreSteenbergen:


veikkoeeva:
I'll link to https://github.com/RehanSaeed/Schema.NET/issues/5.

@AndreSteenbergen This is very interesting. Lurking back to this, how do these changes handle unknown types? To be more clear, my angle is from https://github.com/RehanSaeed/Schema.NET/issues/2#issuecomment-320491715. I've been thinking also something like https://stackoverflow.com/a/42442258/1332416 so that even unknown objects wouldn't be lost, but could be typed later and in the meantime get data stored in `JObject` or somesuch. As noted in https://github.com/RehanSaeed/Schema.NET/issues/5, one could also have a "a fixed function" to normalize away known errors (instead of backing the logic in the core).

A toold to generate POCOs from schema classes is valuable. :)

AndreSteenbergen:
@veikkoeeva What exactly do you mean with unknown types? Microdata and ld+json both "demand" you set the type. I changed the Thing converter to deserialize any thing, returning a strong typed version. Problem is, I did this in a day, so I haven't had the change to fully test this. I did add in an extra object which shouldn't be in there, ItemList, storing Thing that are not really part of the schema.org type, but were given by the document at hand.

AndreSteenbergen:
@veikkoeeva The tool was already made by @RehanSaeed It is part of the source. If it wasn't for that piece of code I would have taken a lot more time.

AndreSteenbergen:
The code is actually possible now, the introduction of the ThingConverter made that possible. everthing will turn out to be the correct strong type.

```
[HttpPost("[action]")]
public IActionResult Test([FromBody] Thing thing)
{
    var e = (Person)thing;
    return Accepted(new Uri("https://wwww.WhereToAskForResults.com"));
}
```

veikkoeeva:
@AndreSteenbergen There might be a case a new type inheriting from `Thing` has been introduced for which there isn't yet a concrete type generated in the library. In this case, would the aforementioned API calls start to fail or would the deserialization succeed? This might require thinking along the lines of https://stackoverflow.com/questions/37076278/is-it-possible-to-deserialize-into-an-expandoobject-or-dictionary-having-stron.

I don't mean to push this point, though. :) I haven't had time to put effort to this myself (priorities, eh), but it looks like while JSON-LD is definitely a way to go, it can also introduce unexpected run-time types one wouldn't want to fail with but perhaps gradually deploy new types into production when necessary (e.g. not just pushing them through the system to storage, but perhaps transform). Also, as you note, one problem is improperly constructed objects. If not failing them, should there be a possibility to introduce converters/fixers (e.g. an extension method or something) that rearranges the types?

These are more like ideas, to add perspectives.

AndreSteenbergen:
I guess the converter I added might fail, check out ThingConverter. I am willing to change the code, if you have a proposed solution. I have other prio's at this moment as well.

AndreSteenbergen:
There is just 1 function that needs to be changed. It checks the built in types. Maybe add an expandoobject there, my personal main focus was getting the microdata parser up and running

RehanSaeed:
Thanks for this PR, there are some good ideas here.

### Unsupported Objects
@veikkoeeva Where there is a new type that Schema.NET does not support, you should still get a `Thing` deserialized correctly since that is the base type of all Schema.org objects. I've raised issue https://github.com/RehanSaeed/Schema.NET/issues/7 to address this. This small issue is up for grabs to be implemented.

### Mixing Types

@AndreSteenbergen Did you think more on the my suggestion [here](https://github.com/RehanSaeed/Schema.NET/issues/5#issuecomment-366685552).

### ASP.NET Core Support

Issue https://github.com/RehanSaeed/Schema.NET/issues/2 addresses this. We could provide a project which uses the HtmlAgilityPack to pull JSON-LD from a HTML page and then deserialize it using Schema.NET. I thought it was better left as documentation as it would only be a few lines of code but looking at the `ThingsParser` in this PR, am I wrong about that @AndreSteenbergen?

AndreSteenbergen:
@RehanSaeed I like the idea. My knowledge about JsonConverters using newtonsoft is not that well. I don't really need to build it myself that often.

The Thingsparser I added was a MicrodataParser. A regular JSON-LD in html is a no-brainer. Get all script tags, with type application/ld+json, and feed the innerText into the deserializer. No pain there. There just aren't any real good implementations to be found to parse Microdata Schema.org items.

AndreSteenbergen:
I can try and work something out using your suggestions next weekend. Prio's ;)

RehanSaeed:
MicrodataParser Interesting...I've not even considered that angle. The code seems pretty involved. What scenarios does it cover or not cover? It should probably be a separate PR/issue.

@AndreSteenbergen I had some code I was working on to rename and refactor `Values<T>` into `OneOrMany<T>` (Also now implements IEnumerable<T>, Count, HasOne, HasMany etc. I'm considering adding ICollection<T> but that makes the struct writable which is probably bad). That's now checked in and merged. I wanted to get this in before you started major rework around `Values<T1...TN>`.

AndreSteenbergen:
Thanks for the heads up, I like that names more than Values ;)

The microdataparser isn't all that complicated, it performs a breadth first search to any node with an itemscope. After that it will try and find nodes with itemprop. The only complex part in there is parsing strings into something we would prefer over string, e.g. decimal, float double etc.

Using the proposed change you did yesterday will make it bit simpler, as you don't need to find classes implementing the interface specified. You would only loop through the T1..Tn trying to figure out the best type to change to string into.

I haven't found any html document it does not parse correctly. The only thing it does not handle correctly are itemrefs, as I didn't build that part, it would probably harder than what is already done.

BTW you are correct, I should have made an extra PR.

AndreSteenbergen:
I do need an Add in the OneOrMany<T>, as with parsing html I don't know in advance the number of items. Most of the time it will be one, but you never know.

psampaio:
If the intent is to have a marker interface as a base for all the alternatives in Values\<Tn\>, keep in mind that, while rare, there are some instances where classes are mixed with enums as alternatives.

For instance, [Offer](https://schema.org/Offer) has an [acceptedPaymentMethod](https://schema.org/acceptedPaymentMethod) that can either be a class ([LoanOrCredit](https://schema.org/LoanOrCredit)) or an enum ([PaymentMethod](https://schema.org/PaymentMethod)).

AndreSteenbergen:
@psampaio it is already in here: for example the paymentmethod you describe is a value of
IAcceptedPaymentMethod which looks like this:
```
IAcceptedPaymentMethod : IWrapper { }
public interface IAcceptedPaymentMethod<T> : IAcceptedPaymentMethod { new T Data { get; set; } }
public class AcceptedPaymentMethodLoanOrCredit : IAcceptedPaymentMethod<LoanOrCredit>
{
SNIPPET
}
public class AcceptedPaymentMethodPaymentMethod : IAcceptedPaymentMethod<PaymentMethod>
{
SNIPPET
}
```

psampaio:
Sorry, I had not seen it on your proposal and had tripped on that myself so thought I should throw it out there. Great work!

RehanSaeed:
@psampaio I think this PR is really doing two things. First it adds a really cool microdata parser and second is that it allows mixing multiple types and solves https://github.com/RehanSaeed/Schema.NET/issues/5. Did you read my comment [here](https://github.com/RehanSaeed/Schema.NET/issues/5#issuecomment-366685552)? I think this will solve the problem of mixed data in a simpler way but I have not had the time to implement it for real.

If we manage to fix the mixed types problem above, it would be great to revisit microdata and would make @AndreSteenbergen's PR a bit simpler too (669 files changed).

AndreSteenbergen:
@RehanSaeed I still haven't had time to change this, I am working with this code for some time now, the only not so nice part is the .Data issue. I am actually in the process of changing all values to interfaces. Using partial classes I can extend the classes I want, even though I am breaking schema.org standards.

RehanSaeed:
@AndreSteenbergen Thats perfectly fine, that's just open source. We all have to decide how much personal time to trade with doing open source for complete strangers. We all struggle with that balance.

RehanSaeed:
See https://github.com/RehanSaeed/Schema.NET/issues/5#issuecomment-492304705 for details of an implementation. Feedback welcome.

PR #12, opened 2018-04-16
Support type array

psampaio:
According to the [spec](https://json-ld.org/spec/latest/json-ld/#specifying-the-type), type can be defined as an array. For C#, this is obviously not valid, so we arbitrarily choose the first one.

Note that this is only valid for sub-properties and not for the main object itself. The reason for that is that the Type property on Thing does not have a custom converter (nor should it?). I'd guess most people will specify the type on their own deserialization call, anyway.

### DIFF
--- a/Source/Schema.NET/ValuesConverter.cs
+++ b/Source/Schema.NET/ValuesConverter.cs
@@ -219,6 +219,22 @@ namespace Schema.NET
         private static string GetTypeNameFromToken(JToken token)
         {
             var typeNameToken = token.Values().FirstOrDefault(t => t.Path.EndsWith("@type"));
+            if (typeNameToken is JArray array)
+            {
+                if (!array.Any())
+                {
+                    throw new ArgumentException("Type was an empty array");
+                }
+
+                typeNameToken.Parent.Remove();
+                if (token is JContainer container)
+                {
+                    var value = array.First.Value<string>();
+                    container.Add(new JProperty("@type", value));
+                    return value;
+                }
+            }
+
             return typeNameToken?.Value<string>();
         }
     }
--- a/Tests/Schema.NET.Test/MultipleTypeTest.cs
+++ b/Tests/Schema.NET.Test/MultipleTypeTest.cs
@@ -1,0 +1,44 @@
+namespace Schema.NET.Test
+{
+    using System;
+    using Newtonsoft.Json;
+    using Xunit;
+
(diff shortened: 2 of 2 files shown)


Source/Schema.NET/ValuesConverter.cs
```diff
@@ -219,6 +219,22 @@ public virtual void WriteObject(JsonWriter writer, object value, JsonSerializer
         private static string GetTypeNameFromToken(JToken token)
         {
             var typeNameToken = token.Values().FirstOrDefault(t => t.Path.EndsWith("@type"));
+            if (typeNameToken is JArray array)
+            {
+                if (!array.Any())
+                {
+                    throw new ArgumentException("Type was an empty array");
```
RehanSaeed: We should try to avoid throwing. It would be better to return a `Thing`.

RehanSaeed:
This is a great. What are the use cases for choosing multiple types? Is this a feature used commonly in the wild? There are a few potential options:

1. Choose the first type.
2. Try to choose a common base type which will probably be `Thing` but might not be.

Perhaps this should be configurable. Perhaps we could use an enum to determine whether to use 1 or 2? What are your thoughts?

Source/Schema.NET/ValuesConverter.cs
psampaio: For sure, I can look into it.

psampaio:
Most of my use cases happen when I'm deserializing json-ld. The input json is usually generated automatically and in this case, given that the property supports multiple values, the result is a list with one item.

One thing to notice, and it applies to your bullet point 2, is that all listed types are equally valid. In the case mentioned in the spec, the properties are valid both for a Restaurant and for a Brewery. To be accurate, you should generate a list of nested entities, one for each type. Something like: get the properties, go to the parent token, remove the original child and replace with a list of two new entities, generated from the set of properties copied initially.

Again, making this work on the root entity would require a large refactor, if possible at all, I'm guessing.

psampaio:
I think your third option matched what I explained, my explanation was just terrible. :)

I agree the best would be to return a collection of object, one for each type. I'll see if I have some time during the weekend to hack a proof-of-concept of this.

RehanSaeed:
See [this](https://github.com/RehanSaeed/Schema.NET/issues/5#issuecomment-492304705) comment for an implementation. Feedback welcome.

PR #16, opened 2018-08-26
Update Breadcrumb unit test to include Id property

icunningham88:
It appears that the current breadcrumb fails the google structured data testing tool found at:

https://search.google.com/structured-data/testing-tool

Updated to include url.


RehanSaeed:
Thanks for your contribution.

Issue #18, opened 2018-10-01
Potential XSS vulnerability


benmccallum:
I think that a lot of people in the ASP.NET MVC world would do something like this to get their schema rendered.

```
<script type="application/ld+json">
    @Model.myThing.ToString();
</script>
```

Problem is with `Json.Convert`'s default settings, that opens up an XSS vulnerability if the data put in a thing's property was untrusted. e.g. a business name on your website could be user-supplied by that business, and they could make their business name `</script><script>alert('gotcha');</script>`.

I think it would be addressed by updating the default Serialization settings in Thing.cs to include `StringEscapeHandling = StringEscapeHandling.EscapeHtml`, but this isn't my area of expertise. That also  might be a breaking change if HTML is allowed in any schema properties?

What do you think? Just opening it up for discussion as trying to avoid this in my own project.

RehanSaeed:
Thanks for the suggestion, I think you are correct in your thinking. Security should be a default. I'd be happy to take a PR with this change. It would be breaking so, it'd be a major version change.

benmccallum:
Sure thing, I'll whip one up when I get a sec.

RehanSaeed:
Will get released soon.

PR #19, opened 2018-10-01
Avoid XSS by escaping HTML during serialization

benmccallum:
Addresses #18


benmccallum:
Damn, tried to avoid cloning locally and just made the change in GitHub, but must have some whitespace. Let me fix that up.

benmccallum:
Fixed that but ran the tests and there's a few that fail as they had HTML in the description property. I can't find any documentation on whether that's valid or not. schema.org just labels it a "Text" field, which doesn't specify further what it can hold, and the examples for some of those types of objects never have HTML, so... should I just strip the HTML tags out?

benmccallum:
Next interesting part is that some tests are failing because `'` is encoded to `\u0027`. Currently investigating how the browser will handle that. If it's like that in the View Source, I can't imagine that's valid for the schema parsing.

RehanSaeed:
I just realized that we don't provide an overload on ToString that accepts a `JsonSerializerSettings` object. We should definately provide that no matter what so that the default behaviour can be overridden.

Maybe we should even go as far as being very explicit and adding a `ToHtmlEncodedString()` method for your use-case. What are your thoughts?

I'll take a look at the problem with the generator tool when I get some time.

benmccallum:
It's tricky. I guess it depends how much we want to impress the security issue on the consumer and also on the usage of this library, i.e. do the majority of people push this value straight into a <script> or do people use it server-to-server for other reasons and not want the html escaping? I'd probably suggest the former, but you may know more about how this is used.

As such, I tend to think a major version bump and enforcing the security by default should be the way to go, so we could introduce an overload that accepts a `JsonSerializationSettings` like you said in case anyone wants the original behaviour or to adjust as they please. This would also mean my updates to the unit tests are still valid, and I could just add a new test for the overload to make sure it uses a user-supplied `JsonSerializationSettings`.

So in summary, I'd:
* Create a `ToString()` overload that allows user-supplied `JsonSerializationSettings`. In it's XML comment, specify to use with caution, and use the escaped html setting unless there's a really good reason not to.
* Update readme to document new default, and option to still customize settings if needed, with caution

benmccallum:
No worries, will fix that up tomorrow. The JSON.NET setting is `StringEscapeHandling = StringEscapeHandling.EscapeHtml`, do you think the method should be called `ToHtmlEncodedString` like you suggested or `ToHtmlEscapedString`? I'm not sure what's more consistent + semantically correct here.

benmccallum:
I don't think they're exactly the same, but I think in this case they same characters as an encode or escape are being handled, so the difference is non-existant. https://stackoverflow.com/questions/4354228/is-encoding-the-same-as-escaping

Finishing everything up now, currently named with escape, so speak now or forever hold your peace ha

RehanSaeed:
Happy with escape in that case.

benmccallum:
Cool, pushed and over to you. Let me know when you get it up on NuGet so I can update my project to use this new built-in method rather than my own. Cheers and thanks for the library!

JamesNK:
> I'm sure @JamesNK probably named it correctly.

image

benmccallum:
> > I'm sure @JamesNK probably named it correctly.
> 
> image

With the most downloaded nuget package on the planet...
were-doomed-gif-9

RehanSaeed:
Ran the tool and updated everything. You'll need to rebase.

benmccallum:
Nice one. Have refreshed from your master. Tests look good. Let's get this released 🎉

README.md
```diff
@@ -26,6 +26,8 @@ The code above outputs the following JSON-LD:
 }
 ```
 
+If writing the result into a `<script>` element, be sure to use the `.ToHtmlEscapedString()` method instead to avoid exposing your website to a Cross-Site Scripting attack. See the [example below](#important-security-notice).
```
RehanSaeed: Nice.

Source/Schema.NET/Thing.Partial.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Schema.NET
+namespace Schema.NET
```
RehanSaeed: Why is it saying something changed on this line. Weird.

Tests/Schema.NET.Test/JobPostingTest.cs
```diff
@@ -10,7 +10,7 @@ public class JobPostingTest
         private readonly JobPosting jobPosting = new JobPosting()
         {
             Title = "Fitter and Turner", // Required
-            Description = "<p>Widget assembly role for pressing wheel assemblies.</p><p><strong>Educational Requirements:</strong> Completed level 2 ISTA Machinist Apprenticeship.</p><p><strong>Required Experience:</strong> At least 3 years in a machinist role.</p>", // Required
+            Description = "Widget assembly role for pressing wheel assemblies. Educational Requirements: Completed level 2 ISTA Machinist Apprenticeship. Required Experience: At least 3 years in a machinist role.", // Required
```
RehanSaeed: Any reason to remove the HTML tags here?

Source/Schema.NET/Thing.Partial.cs
benmccallum: that is weird. can't say i know why. if it was whitespace on the end it would've highlighted i thought.

Tests/Schema.NET.Test/JobPostingTest.cs
benmccallum: Originally did this as my first changes had made the default `.ToString()` use the HTML escaping. As such I would've had to encode these tags. I think encoding would work, but I couldn't find any examples on schema.net of html tags in or being allowed in these properties, so I just removed. Should be able to add back though if you'd like and we could test the encoding with the Google schema tester tool.

RehanSaeed: I think I remember copying this data from a Google website, probably the schema testing tool. It would be nice to match up with it to make sure everything is working with Google, since that is the main use case for this. Very minor change, then we can merge and release.

PR #21, opened 2018-10-18
Reverting removal of HTML tag in this test.

benmccallum:
As per discussion elsewhere. Putting this back so then we can a release out 👍 . Let me know when it's up on NuGet so I can update my proj to latest. Ta.


RehanSaeed:
Thanks added to the contributions section of the home page:

PR #27, opened 2018-11-28
Add .Net Framework 4.5.2 target

x2764tech:
I'd really like to use this project in an Umbraco site, but Umbraco is `net452` only. 
This change should allow targeting .Net Framework 4.5.2.


RehanSaeed:
This project supports `netstandard1.1` and `netstandard2.0` which means that .NET 4.5.2 is already supported.

Feel free to reopen if I've missed something.

PR #30, opened 2019-02-05
Added linux support for tool

AndreSteenbergen:
Added the directory separator, instead of hard coded back slash

### DIFF
--- a/Source/Schema.NET.Tool/Services/SchemaService.cs
+++ b/Source/Schema.NET.Tool/Services/SchemaService.cs
@@ -1,6 +1,7 @@
 namespace Schema.NET.Tool.Services
 {
     using System;
+    using System.IO;
     using System.Collections.Generic;
     using System.Linq;
     using System.Text;
@@ -102,7 +103,7 @@ namespace Schema.NET.Tool.Services
                             Description = classDescription,
                             Id = new Uri($"http://CombinedClass/{className}"),
                             IsCombined = true,
-                            Layer = @class.IsCombined ? @class.Layer : $"{@class.Layer}\\combined",
+                            Layer = @class.IsCombined ? @class.Layer : $"{@class.Layer}{Path.DirectorySeparatorChar}combined",
                             Name = className
                         };
                         combinedClass.Properties = @class
@@ -200,7 +201,7 @@ namespace Schema.NET.Tool.Services
             return new Enumeration()
             {
                 Description = schemaClass.Comment,
-                Layer = schemaClass.Layer + "\\enumerations",
+                Layer = $"{schemaClass.Layer}{Path.DirectorySeparatorChar}enumerations",
                 Name = schemaClass.Label,
                 Values = values
             };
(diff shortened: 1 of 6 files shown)


Source/Schema.NET/core/WebPageElement.cs
```diff
@@ -15,5 +15,19 @@ public partial class WebPageElement : CreativeWork
         /// </summary>
         [DataMember(Name = "@type", Order = 1)]
         public override string Type => "WebPageElement";
+
+        /// <summary>
+        /// A CSS selector, e.g. of a &lt;a class="localLink" href="http://schema.org/SpeakableSpecification"&gt;SpeakableSpecification&lt;/a&gt; or &lt;a class="localLink" href="http://schema.org/WebPageElement"&gt;WebPageElement&lt;/a&gt;. In the latter case, multiple matches within a page can constitute a single conceptual "Web page element".
+        /// </summary>
+        [DataMember(Name = "cssSelector", Order = 206)]
+        [JsonConverter(typeof(ValuesConverter))]
+        public OneOrMany<CssSelectorType>? CssSelector { get; set; }
```
RehanSaeed: I guess that you did not run the tool to generate these two properties. Maintaining this by hand is not really scalable due to the number of types. However, the tool project has some overrides that add extra properties that don't exist in schema.org but are defined in Google structured data.

Source/Schema.NET.Tool/Services/SchemaService.cs
```diff
@@ -102,7 +103,7 @@ private static void CombineMultipleParentsIntoCombinedClass(List<Class> classes)
                             Description = classDescription,
                             Id = new Uri($"http://CombinedClass/{className}"),
                             IsCombined = true,
-                            Layer = @class.IsCombined ? @class.Layer : $"{@class.Layer}\\combined",
+                            Layer = @class.IsCombined ? @class.Layer : $"{@class.Layer}{Path.DirectorySeparatorChar}combined",
```
RehanSaeed: This is all good stuff and I can merge this if it was a separate PR.

Source/Schema.NET/core/WebPageElement.cs
AndreSteenbergen: I actually ran the tool as is, and these came up automatically. I actually added those 2 files by hand, as these were not generated (they are basically strings)

AndreSteenbergen:
This is actually just a phase 1 commit; a larger project is coming to allow mixed types, without adding a whole bunch of interfaces (you can check out Values<T1>). I think it won't be that hard. just that jsonconvertor ..... :( .

Source/Schema.NET/core/WebPageElement.cs
RehanSaeed: See https://github.com/RehanSaeed/Schema.NET/tree/master/Source/Schema.NET.Tool/Overrides for examples of how to add stuff like this.

RehanSaeed:
Small PR's make things easier to merge. The directory separator can be a quick and easy merge.

RehanSaeed:
Thanks for your work by the way!

AndreSteenbergen:
Will make a PR that is smaller, and also check out that link thx

AndreSteenbergen:
I'll close this one, and reopen another, easiest for me.

PR #31, opened 2019-02-06
Added support for linux in the tooling

AndreSteenbergen:
Added support for linux in the tooling by changing the directory separator from hardcoded backslash to {Path.DirectorySeparatorChar}


RehanSaeed:
Thanks for that.

PR #32, opened 2019-02-07
Pull request to discuss mixing types

AndreSteenbergen:
- open to all suggestions ...

- keep implicit operators intact as much as possible
- be sure no invalid types get into object
- be able to mix types when necessary

- Serialization should be in extra project, not everyone needs to serialize this data.


AndreSteenbergen:
Persons, Brands etc I regularly come across these on websites without specifying itemtype, only a string, it's quite valid imho to define <book><author>Some dude</author></book> or <car><brand>volvo</brand></car> is Volvo a brand(Brand) or brand(Organization) .....

Tests/Schema.NET.Test/ValuesTest.cs
```diff
@@ -0,0 +1,57 @@
...
+    using System;
+    using Newtonsoft.Json;
+    using Xunit;
+
+    public class ValuesTest
+    {
+        [Fact]
+        public void CanSetValuesIntUsingString()
+        {
+            Values<int> vInt = (Values<int>)"someTest";
```
RehanSaeed: I'm not sure I get this yet but why accept `string` for every type?

AndreSteenbergen: I actually removed that ;)

But I am having some issues with parsing data. I had some sites running a brand, and it was specified as brand="XXXXX". Brand can not have string, needs to be strong type Brand or Organization. Also I am wondering about author="XXX" for creativework.

It didn't feel ok, so I removed it again. I do need to think about types string to brand, string to author etc. But I don't think that should be done in the schema.org classes. If there are any ideas on how to handle those hard cases where string is not allowed, I actually don't want to loose data, because it doesn't fit the objects.

At this moment I am more leanings against an extra Dictionary in Thing. Containing all stuff that doesn't fit where it is supposed to fit. Don't know about that one either ..... schema.org is too big to manually work out all string -> object conversions (at least I think so)

RehanSaeed:
See [this](https://github.com/RehanSaeed/Schema.NET/issues/5#issuecomment-492304705) comment for an implementation. Feedback welcome.

Issue #29, opened 2018-12-06
OneOrMany ctor to accept null?


benmccallum:
Just ran into a situation where I'm passing in `null` into a property that's `OneOrMany<T>` and the ctor(T item) throws an exception. 

Just wondering whether this is necessary or we can allow null to be set and handled internally. Otherwise code like below becomes much more complicated as I need to null-check every assignment and can't use object initialization, instead need to instantiate then assign properties if not null.

```
            return new LocalBusiness
            {
                Name = Name,
                Image = Image,
                PriceRange = "$$",
                Address = new PostalAddress
                {
                    StreetAddress = Address,
                    AddressLocality = Suburb,
                    AddressRegion = State?.ToUpper(),
                    PostalCode = Postcode
                }
            };
```

What do you think? Would it be possible given all the generics/internals in play here?

As is the behaviour for other properties that allow `null`, I'd expect if `OneOrMany<T>` was set to null to not render out to JSON. Or I guess respect the `JsonSerializerSettings` null handling we now have an overload for. Assuming the value returned is null that should all work ootb.

benmccallum:
I feel like this null check and throw might be new. Suprised I didn't get it originally.

benmccallum:
Nope, checked git blame and commit is from Feb. Hmm, probably a few pages broken on my site then ha!

RehanSaeed:
I'm assuming you mean that we just remove the null check at https://github.com/RehanSaeed/Schema.NET/blob/09b2616a253d329c24134da9d5d68ae69a742176/Source/Schema.NET/OneOrMany%7BT%7D.cs#L23-L26

I.....guess thats ok. Not sure if there would be any unintended consequences of this.

benmccallum:
Yea that's the one. It's super tedious to handle nulls as it is right now. I can maybe give it a go removing it and seeing what happens if you're open to the idea?

RehanSaeed:
Please do.

tompazourek:
I had the same issue, the site broke when I updated to the latest version.

I resolved it temporarily by appending ` ?? new OneOrMany<string>?()`, e. g.:

```
return new LocalBusiness
{
    Name = Name,
    Image = Image,
    PriceRange = "$$",
    Address = new PostalAddress
    {
        StreetAddress = Address,
        AddressLocality = Suburb,
        AddressRegion = State?.ToUpper() ?? new OneOrMany<string>?(),
        PostalCode = Postcode
    }
};
```

[...]

PR #36, opened 2019-03-08
Proposed fix for null value handling (#29)

nickevansuk:
This project is really great! We've used it as the basis for [OpenActive.NET](https://www.nuget.org/packages/OpenActive.NET/) - which builds on Schema.NET to create a client library for the [OpenActive](https://www.openactive.io/) Initiative. The OpenActive [model](https://www.w3.org/2017/08/modelling-opportunity-data/) builds on Schema.org, so building on this library was the perfect fit. OpenActive.NET references the Schema.NET NuGet package and we were able to do everything using extensions and inheritance alone, which is ideal - great work making such an extensible framework!

We'd love to continue to contribute to Schema.NET as OpenActive evolves, as it looks like our objectives are completely aligned here.

The only issue we've had is that we needed to fix https://github.com/RehanSaeed/Schema.NET/issues/29 to make Schema.NET work for OpenActive data publishers, who frequently have null values in their source data.

For null handling OpenActive has some specific requirements (which apply to Schema.org JSON-LD too, though they are not _required_ there), when serialising:
- Nulls _must_ be ignored
- Empty arrays _must_ be ignored
- Strings that are empty or contain whitespace _must_ be ignored.

For comparison with schema.org: although the last two in the context of schema.org just make for better structured data (Google won't do anything with empty arrays or blank strings, so there is no scenario where they would be included in data intentionally), references for ignoring nulls are part of the JSON-LD spec:
> A key-value pair in the body of a JSON-LD document whose value is null has the same meaning as if the key-value pair was not defined. - [JSON-LD spec](https://json-ld.org/spec/latest/json-ld/#terminology)

Hence we'd be keen to propose that all three of the above requirements are also baked into Schema.NET :)

Back to https://github.com/RehanSaeed/Schema.NET/issues/29, following some investigation there seems to be a little more to this, especially if we want to ignore null values in all circumstances... As soon as we create a non-default struct value, Json.NET will not allow the value to be hidden easily even if properties of that value are null. If the `ValuesConverter` uses `writer.WriteNull()` when rendering the value, it will ignore the setting of `NullValueHandling.Ignore` (which is applied before `ValuesConverter.WriteJson` is even called). Unless we want to unwind during JSON writing ([messy](https://stackoverflow.com/questions/33241039/nullvaluehandling-ignore-with-jsonconverterwritejson)) we need to prevent even creating the non-default struct in the first place. To complicate this further even a default struct value is not recognised by `DefaultValueHandling.Ignore`, which means it looks like we need to switch back to classes from structs.

The PR here does exactly this, and keeps the existing null checks in place in the constructors to ensure that IValue classes with null values aren’t accidentally created. It also improves the `EventTest` to cover the various null and empty scenarios.

Note the CI errors as this PR does not include the regenerated model for readability (the regenerated model just removes the Nullable shorthand `T?` due to the struct->class change). The regenerated model can be found in our fork (diff [here](https://github.com/RehanSaeed/Schema.NET/compare/master...openactive:master)) for reference. 

The main downside I can see is any performance impact of switching from structs to classes, which doesn't seem to have a significant effect on performance from a quick test. What testing have you done on structs vs. classes in the scenario here? Do you think we need to do more or given the average size of a Schema.org document (small for structured data) would you be happy to accept a PR on this issue without this?

Look forward to your thoughts,

Nick

### DIFF
--- a/Source/Schema.NET.Tool/ViewModels/Property.cs
+++ b/Source/Schema.NET.Tool/ViewModels/Property.cs
@@ -38,7 +38,7 @@ namespace Schema.NET.Tool.ViewModels
 
             var adjustedTypes = string.Join(", ", this.Types.Select(x => x.CSharpTypeString));
             var rootType = this.Types.Count == 1 ? "OneOrMany" : "Values";
-            var typeString = $"{rootType}<{adjustedTypes}>?";
+            var typeString = $"{rootType}<{adjustedTypes}>";
 
             stringBuilder.AppendIndentLine(indent, $"[DataMember(Name = \"{this.JsonName}\", Order = {this.Order})]");
 
--- a/Source/Schema.NET/OneOrMany{T}.cs
+++ b/Source/Schema.NET/OneOrMany{T}.cs
@@ -9,13 +9,13 @@ namespace Schema.NET
     /// </summary>
     /// <typeparam name="T">The type of the values.</typeparam>
     /// <seealso cref="ICollection{T}" />
-    public struct OneOrMany<T> : IEnumerable<T>, IValue
+    public class OneOrMany<T> : IEnumerable<T>, IValue
     {
         private readonly List<T> collection;
         private readonly T item;
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="OneOrMany{T}"/> struct.
+        /// Initializes a new instance of the <see cref="OneOrMany{T}"/> class.
         /// </summary>
         /// <param name="item">The single item value.</param>
         public OneOrMany(T item)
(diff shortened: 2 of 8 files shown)


RehanSaeed:
To support mixed types, a lot has changed in a potential new version. Please see [this](https://github.com/RehanSaeed/Schema.NET/issues/5#issuecomment-492304705) comment for details. Feedback welcome.

nickevansuk:
Looks good @RehanSaeed, though it doesn't seem to fix this particular issue. Need to fix the implicit operators to fix this.

If you let me know your thoughts on https://github.com/RehanSaeed/Schema.NET/issues/29#issuecomment-473652862 I can bring this PR up-to-date with master again

RehanSaeed:
Give I now have a fix in the `mixed-types` branch, can this be closed?

PR #41, opened 2019-04-25
Test for Book.Author typeof Person

radetskyi:


RehanSaeed:
Thanks! This was really helpful. I've fixed it in the latest commit. Even after fixing the test was not quite right, so won't merge this.

RehanSaeed:
Thanks! This was really helpful. I've fixed it in the latest commit. Even after fixing the test was not quite right, so won't merge this.

PR #48, opened 2019-05-09
Fix to handle Date-types from Schema.org

MEmanuelsson:
Adds support for date in ISO 8601 format

- YYYY eg. 2015
- ~~YYYY-MM eg. 2015-02~~
- YYYY-MM-DD eg. 2015-02-28

Probably a good idea to also create a special struct for DateTime-handling (replacing usage of DateTimeOffset) but haven't done that at this moment. Let me know if you want me to include that also in a PR?


RehanSaeed:
There are some good ideas in this PR. However, I think there are a few problems with the `IsoDate` type:
  - The name is misleading. `DateTime` and `DateTimeOffset` also get serialized to ISO8601. Perhaps it could have been named just `Date`.
  - It's a `class` instead of a `struct`.
  - I don't think we need to create a new `IsoDate` type. I think we can get away with repurposing `DateTime` and then using a custom `JsonConverter` as you have done to control the output of year, month or day. Is that sensible or am I missing something?

MEmanuelsson:
Actually renamed it from Date to IsoDate (to avoid the conflict with the reserved word) in the last minute.  

The thing I don't understand when you say that we should be able to use DateTime instead of a special struct (or class which I ended up doing here, sorry) is how we can new up a DateTime with only a year as input? Doing that with new DateTime(2019, 0, 0) for instance results in ArgumentOutOfRangeException "Year, Month, and Day parameters describe an un-representable DateTime."

RehanSaeed:
> how we can new up a DateTime with only a year as input?

This is what I was missing. yes `new DateTime(2019, 0, 0)` is not ideal. I guess it's a tradeoff. Repurposing `DateTime` has some advantages because it's so ubiquitous but this is it's downside.

What if we also allowed `int` for these occasions, so you can enter a year on it's own? Would that solve this issue?

MEmanuelsson:
> What if we also allowed int for these occasions, so you can enter a year on it's own? Would that solve this issue?

That would solve the problem that I faced in the particular case of a car's model year, however it doesn't solve the occassions when you might want to set the "date" to a combination of year and month.

RehanSaeed:
How about this for a start then:

```
        public struct Date : IComparable, IComparable<Date>, IConvertible, IEquatable<Date>, IFormattable
        {
            private readonly DateTime dateTime;

            public Date(int year) => this.dateTime = new DateTime(year, 0, 0);

            public Date(int year, int month) => this.dateTime = new DateTime(year, month, 0);

            public Date(int year, int month, int day) => this.dateTime = new DateTime(year, month, day);

            public int Year => this.dateTime.Year;

            public int Month => this.dateTime.Month;

            public int Day => this.dateTime.Day;

            public int DayOfYear => this.dateTime.DayOfYear;

            public DayOfWeek DayOfWeek => this.dateTime.DayOfWeek;

            public static implicit operator Date(DateTime dateTime) =>
                new Date(dateTime.Year, dateTime.Month, dateTime.Day);

            public static implicit operator DateTime(Date date) => date.dateTime;
        }
```

We just wrap `DateTime` but still provide implicit operators to convert to and from `DateTime`. I'll leave implementing all the interfaces as an exercise for you.

MEmanuelsson:
I can give it a go but will this code not give the exact problem as I mentioned earlier (ArgumentOutOfRangeException "Year, Month, and Day parameters describe an un-representable DateTime."), 

`public Date(int year) => this.dateTime = new DateTime(year, 0, 0);`

RehanSaeed:
In that case, perhaps we have more than one `struct`? `Year`, `YearMonth` and `Date`? Thoughts?

MEmanuelsson:
I think that's a design decision that you should have the final saying about. Either go for:

1. Custom `Date` struct to handle year, year-and-month and date, together with .NET's `DateTime `to handle date-with-time
2. Custom `Year`, `YearMonth` and `Date` structs, together with .NET's `DateTime` to handle date-with-time

I think I'm more in favor for option 1 since that then also corresponds better with how schema.org separate between Date (https://schema.org/Date) and DateTime (https://schema.org/DateTime)

RehanSaeed:
I was thinking about this over lunch. I think we can get 2/3 by adding `int` and `DateTime` support which I think we should do no matter what. Only thing missing is year and month support. Is that something you really need?

MEmanuelsson:
For me personally, I have no need for a combination of year and month, at the moment, but since it's specified by schema.org that this is something that they accept based on the ISO 8601 standard, maybe someone else using this library now or in the future has/will have this need? 

I feel that the PR that I've provided (with some modifications based on your comments) supports 3/3 but if you want to go for a switch from `DateTimeOffset` to a mix of `int` and `DateTime` to support 2/3, it's your decision.

RehanSaeed:
With or without your PR, we should support `int` and `DateTime` because those are built in types. In theory that should also be a simple PR too. I say we do those and if someone really wants year and month we can add it at a later date. How does that sound?

MEmanuelsson:
Sure, let's do that. Do you want me to create a new PR for this or can I just continue on this one? Not too experienced with handling pull requests to open source projects :-)

RehanSaeed:
Whatever is easier for you.

MEmanuelsson:
Could use a bit of your help to point me in the right direction regarding this PR, so that I don't break any design decisions that are fundamental in the project. I've so far looked at two approaches to add both `int` and `DateTime` to properties of type Date from Schema.org:

- Update the function GetCSharpTypeString in SchemaService.cs to return a List of strings instead of just a string. Since that will have impact on quite a lot of places in the code I'd like to have a OK from you before I proceed further with this.

- Create a class similar to AddTextTypeToActionTarget for the Date-class. Haven't gotten this to work all the way for this primitive class but could probably look into this further if you think this is a good approach. 

Or am I missing something obvious, some obvious and optimal way to implement this change? How would you go ahead in adding this if you were to do it (in pseudo code)?

RehanSaeed:
Go for the first option, it's the right approach.

MEmanuelsson:
Thanks for the input! 

So I think I have a working solution for this now but I feel that I should also remove the usage of `DateTimeOffset` and instead use `DateTime` also here (using custom JsonConverter to render the correct value based on input). For instance one problem comes with fields that can be either Date or DateTime in Schema.org, like startDate on the type Role(https://schema.org/Role) which is generated like this now: 

        /// <summary>
        /// The start date and time of the item (in &lt;a href="http://en.wikipedia.org/wiki/ISO_8601"&gt;ISO 8601 date format&lt;/a&gt;).
        /// </summary>
        [DataMember(Name = "startDate", Order = 207)]
        [JsonConverter(typeof(DateTimeToIso8601DateValuesConverter))]
        public Values<int?, DateTime?, DateTimeOffset?>? StartDate { get; set; }

It should be enough with only `int `and `DateTime `as options here, thoughts regarding this?

RehanSaeed:
If schema.org says a property is both a Date and DateTime then we should support `int`, `DateTime` and `DateTimeOffset`.

If schema.org says a property is a Date only then we should support `int` and `DateTime`. Right now the `GetCSharpTypeString` method only returns one type but I think it should return a collection of types.

MEmanuelsson:
If you look at my latest commit the function `GetCSharpTypeString` now returns a list of strings. Feel free to dig deeper into this commit and get back to me with issues you might still find. 

In the JsonConverver `DateTimeToIso8601DateValuesConverter` I do handle the possibility that a user adds information about time also in a `DateTime` object, perhaps I should remove that then if `DateTimeOffset` is supposed to be used for `DateTime` from Schema.org?

RehanSaeed:
Yes, I think that should be removed as DateTime is targeted purely at shema.org Date's.

MEmanuelsson:
Pushed update where time-handling is removed

RehanSaeed:
Thank you @MEmanuelsson for all your hard work. Your changes diverged a bit from master, so I've rebased your changes and commited them (I was working on other PR's and changes myself to add support for mixed types, and various other things). I'll get them released in a major 4.0.0 update. As thanks, I'll also add you to the front page as a contributor to the project.

MEmanuelsson:
My pleasure! Happy that I could help!

PR #50, opened 2019-05-17
Updated test to cover null checks for #29

nickevansuk:
This PR includes the specifically enhanced test from https://github.com/RehanSaeed/Schema.NET/pull/36 that covers the issue outlined in https://github.com/RehanSaeed/Schema.NET/issues/29, specifically for testing nullable values against the [mixed-types](https://github.com/RehanSaeed/Schema.NET/tree/mixed-types) branch.


RehanSaeed:
Thanks for that!

PR #53, opened 2019-05-27
Add support for IsNullOrWhiteSpace and improve tests

nickevansuk:
@RehanSaeed as discussed at https://github.com/RehanSaeed/Schema.NET/issues/29, please see updated IsNullOrWhiteSpace via implicit operators.

Everything works great for OneOrMany<>, however additional tests for Values<> have revealed that https://github.com/RehanSaeed/Schema.NET/issues/29 is only fixed for properties that use OneOrMany<> and not for properties that use Values<>.

See failing tests in this PR for examples.

### DIFF
--- a/Source/Schema.NET/OneOrMany{T}.cs
+++ b/Source/Schema.NET/OneOrMany{T}.cs
@@ -107,7 +107,7 @@ namespace Schema.NET
         /// <param name="item">The single item value.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
-        public static implicit operator OneOrMany<T>(T item) => item != null && item.GetType() == typeof(string) && string.IsNullOrWhiteSpace(item as string) ? default : new OneOrMany<T>(item);
+        public static implicit operator OneOrMany<T>(T item) => item != null && IsStringNullOrWhiteSpace(item) ? default : new OneOrMany<T>(item);
 #pragma warning restore CA2225 // Operator overloads have named alternates
 
         /// <summary>
@@ -116,7 +116,7 @@ namespace Schema.NET
         /// <param name="array">The array of values.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
-        public static implicit operator OneOrMany<T>(T[] array) => new OneOrMany<T>(array);
+        public static implicit operator OneOrMany<T>(T[] array) => new OneOrMany<T>(array?.Where(x => x != null && !IsStringNullOrWhiteSpace(x)));
 #pragma warning restore CA2225 // Operator overloads have named alternates
 
         /// <summary>
@@ -125,7 +125,7 @@ namespace Schema.NET
         /// <param name="list">The list of values.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
-        public static implicit operator OneOrMany<T>(List<T> list) => new OneOrMany<T>(list.Where(x => x != null && !(x.GetType() == typeof(string) && string.IsNullOrWhiteSpace(x as string))));
+        public static implicit operator OneOrMany<T>(List<T> list) => new OneOrMany<T>(list?.Where(x => x != null && !IsStringNullOrWhiteSpace(x)));
 #pragma warning restore CA2225 // Operator overloads have named alternates
 
         /// <summary>
@@ -272,5 +272,7 @@ namespace Schema.NET
 
             return 0;
         }
+
+        private static bool IsStringNullOrWhiteSpace(T item) => item.GetType() == typeof(string) && string.IsNullOrWhiteSpace(item as string);
     }
 }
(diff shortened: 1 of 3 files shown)


Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -107,7 +107,7 @@ public int Count
         /// <param name="item">The single item value.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
-        public static implicit operator OneOrMany<T>(T item) => new OneOrMany<T>(item);
+        public static implicit operator OneOrMany<T>(T item) => item != null && item.GetType() == typeof(string) && string.IsNullOrWhiteSpace(item as string) ? default : new OneOrMany<T>(item);
```
RehanSaeed: If you know it's a string, you can just cast it instead of `item as string`.

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -125,7 +125,7 @@ public int Count
         /// <param name="list">The list of values.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
-        public static implicit operator OneOrMany<T>(List<T> list) => new OneOrMany<T>(list);
+        public static implicit operator OneOrMany<T>(List<T> list) => new OneOrMany<T>(list.Where(x => x != null && !(x.GetType() == typeof(string) && string.IsNullOrWhiteSpace(x as string))));
```
RehanSaeed: Same here.

Also, this code was relying on the constructor to throw if the list was null but we can't do that anymore, so you'll need to check for null and throw ArgumentNullException.

Tests/Schema.NET.Test/OneOrManyTest.cs
```diff
@@ -237,5 +237,50 @@ public void GetEnumerator_OneItem_ReturnsOne()
         [Fact]
         public void GetHashCode_TwoItems_HashCodeEqualToTwoItems() =>
             Assert.Equal(NET.HashCode.OfEach(new List<int>() { 1, 2 }), new OneOrMany<int>(1, 2).GetHashCode());
+
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        [InlineData(" ")]
+        [InlineData("  ")]
+        public void ToString_EmptyOrWhiteSpace_BookOmitsNameProperty(string name)
```
RehanSaeed: Can we rename `EmptyOrWhiteSpace` to `NullEmptyOrWhiteSpace`. Very minor.

RehanSaeed:
Also, some of the tests seem to be failing. I literally just added Windows and Ubuntu support for AppVeyor and Azure DevOps, so that came in handy.

RehanSaeed:
Finally, one other thing. I added the following new implicit operator:

```
public static implicit operator OneOrMany<T>(T[] array) => new OneOrMany<T>(array);
```

We'd want to do the same with this.

nickevansuk:
Great, no problem, will make the changes you suggest.

What should we do about the failing tests for properties that use `Values<>`?

It looks like https://github.com/RehanSaeed/Schema.NET/issues/29 was only fixed for properties that use `OneOrMany<T>` and not a property uses `Values<>`.

See this test for an example:
https://github.com/nickevansuk/Schema.NET/blob/master/Tests/Schema.NET.Test/core/EventTest.cs#L43

Source/Schema.NET/OneOrMany{T}.cs
nickevansuk: Apparently not with generics, if I switch to `(string)item` I get a compiler error of `Cannot convert type 'T' to 'string'`

RehanSaeed:
With regards to null `Values<>`, I think that should be a separate PR. I'm experimenting with removing the `?` on [this](https://github.com/RehanSaeed/Schema.NET/blob/131cf9fa343b495faac7287053ac78536e82fd8e/Source/Schema.NET.Tool/ViewModels/Property.cs#L38) line in the same way we did for `OneOrMany`. Would be great if you had a chance to also take a look as my spare time is running low.

Source/Schema.NET/OneOrMany{T}.cs
nickevansuk: Doesn't look like you can raise an exception within an implicit operator method. Using null propagation works though (via `list?.Where(...)`), and probably cleaner

RehanSaeed: Cool.

nickevansuk:
Ok agreed, let's split that out into a separate PR, I'll remove the tests from this PR in that case, so that it passes.

```

        [Theory]
        [InlineData(null)]
        [InlineData("")]
        [InlineData(" ")]
        [InlineData("  ")]
        public void ToString_NullEmptyOrWhiteSpace_OrganizationOmitsAddressProperty(string address)
        {
            var organization = new Organization() { Address = address };
            Assert.Equal("{\"@context\":\"http://schema.org\",\"@type\":\"Organization\"}", organization.ToString());
        }
```

and 

` new Offer { Category = NullString,` in EventTest.cs

nickevansuk:
All tests pass on my side, so hopefully should be good to merge?

nickevansuk:
Looks like it's the Ubuntu image failing, but otherwise it passes

RehanSaeed:
Thanks for this. Merged.

PR #54, opened 2019-05-27
First attempt fixing null properties of Values<>

nickevansuk:
@RehanSaeed I've had a go at refactoring along the lines you explained, but all nulls now seem to be rendering... for both `OneOrMany<>` and `Values<>`!

Did I miss something here?


nickevansuk:
Closing to re-submit based on newly merged master...

RehanSaeed:
Yeah sorry about that. I was seeing if I could get tests to run under net461 under ubuntu + mono. Very tricky.

nickevansuk:
Sounds like it!!

PR #55, opened 2019-05-27
First attempt fixing null properties of Values<>

nickevansuk:
@RehanSaeed I've had a go at refactoring along the lines you explained, but all nulls now seem to be rendering... for both `OneOrMany<>` and `Values<>`!

Did I miss something here?

### DIFF
--- a/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
+++ b/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
@@ -143,6 +143,11 @@ namespace Schema.NET
         public bool HasValue4 { get; }
 
         /// <summary>
+        /// Gets whether the value of either type <typeparamref name="T1" />, type <typeparamref name="T2" />, type <typeparamref name="T3" />, or type <typeparamref name="T4" /> has a value.
+        /// </summary>
+        public bool HasValue => this.HasValue1 || this.HasValue2 || this.HasValue3 || this.HasValue4;
+
+        /// <summary>
         /// Gets the value of type <typeparamref name="T1" />.
         /// </summary>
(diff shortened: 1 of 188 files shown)


Source/Schema.NET/Values{T1,T2,T3,T4}.cs
```diff
@@ -142,6 +142,11 @@ public Values(IEnumerable<object> items)
         /// </summary>
         public bool HasValue4 { get; }
 
+        /// <summary>
+        /// Gets whether the value of either type <typeparamref name="T1" />, type <typeparamref name="T2" />, type <typeparamref name="T3" />, or type <typeparamref name="T4" /> has a value.
+        /// </summary>
+        public bool HasValue => this.HasValue1 || this.HasValue2 || this.HasValue3 || this.HasValue4;
```
RehanSaeed: Nice idea. This should be another PR. Looks like we might not get this one in quickly.

RehanSaeed:
Yes, I tried the same and got the same weird result. Needs more investigation.

Source/Schema.NET/Values{T1,T2,T3,T4}.cs
nickevansuk: Ye this was mainly so the tests would continue passing, as quite a few relied on Nullable<>.HasValue, and broke when we removed the `?`

nickevansuk:
Just to note that I'm on leave for the next few days, but any ideas from anyone on this very welcome!

RehanSaeed:
Added your `HasValue` idea to master.

RehanSaeed:
Closing in favour of https://github.com/RehanSaeed/Schema.NET/pull/84. Thanks for your efforts!

PR #67, opened 2019-06-18
ValuesJsonConverter.cs refactoring

kirk-marple:
@RehanSaeed I've refactored ValuesJsonConverter.cs a good bit.

This passes all unit tests, and even better, doesn't hit your exception 'catch' statements for any of the unit tests. Note some of the // REVIEW comments.

I may have taken liberties with reusing common refactored methods between the single type, and multiple type arg, use cases, but it seems to work.

NOTE: I've kept same semantics where it picks the *last* matching type in the list, by going right-to-left and selecting first one that matches.  This should be a perf benefit, since it was doing wasted work before.

### DIFF
--- a/Source/Schema.NET/ValuesJsonConverter.cs
+++ b/Source/Schema.NET/ValuesJsonConverter.cs
@@ -163,7 +163,7 @@ namespace Schema.NET
 
         private static object ParseTokenArguments(JToken token, JsonToken tokenType, Type type, object value)
         {
-            const int SCHEMA_ORG_LENGTH = 18;
+            const int SCHEMA_ORG_LENGTH = 18; // equivalent to "http://schema.org/".Length
             object args = null;
             var unwrappedType = type.GetUnderlyingTypeFromNullable();
             if (unwrappedType.GetTypeInfo().IsEnum)
--- a/Source/Schema.NET/ValuesJsonConverter.cs
+++ b/Source/Schema.NET/ValuesJsonConverter.cs
@@ -45,44 +45,20 @@ namespace Schema.NET
             object argument = null;
 
             var tokenType = reader.TokenType;
-            var value = SanitizeReaderValue(reader, tokenType);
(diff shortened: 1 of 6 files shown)


Source/Schema.NET/ValuesJsonConverter.cs
```diff
@@ -212,6 +161,216 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
         public virtual void WriteObject(JsonWriter writer, object value, JsonSerializer serializer) =>
             serializer.Serialize(writer, value);
 
+        private static object ParseTokenArguments(JToken token, JsonToken tokenType, Type type, object value)
+        {
+            const int SCHEMA_ORG_LENGTH = 18;
```
RehanSaeed: Can we please add a comment about what this length is?

kirk-marple: Sure, I just flattened that "inline string".Length into a constant.

kirk-marple:
> This looks really promising. Thank you for your contribution. Are there any further tests we could add to make sure that this is working as expected?

I'll put some thought into this.  Have to compare all possible combinations of values/types and make sure we have tests to cover them.

I'm using MusicRecording, MusicGroup, MusicEvent, VideoObject, MusicVenue (and just Thing) in my own app, and not seeing it hit any exceptions with my use cases either.

I think the big change was ignoring string to interface/class casting; that was causing a lot of the exceptions.

kirk-marple:
@RehanSaeed Just added tests for MusicRecording, MusicGroup, MusicEvent, MusicVenue, MusicAlbum.  All passed w/o any further changes.  And added the comment you suggested.  I'd say this is good to go.

RehanSaeed:
Thank you for your contribution. Added you to the front page.

RehanSaeed:
Also thanks for the extra tests, they are beautiful.

PR #84, opened 2019-09-05
fixing null Values<T>

shervinw:
Based from the original [null Values<T> fix attempt](https://github.com/RehanSaeed/Schema.NET/pull/55) from @nickevansuk, this PR just adds minor changes to solve the problem. Also see the [problem origin](https://github.com/RehanSaeed/Schema.NET/issues/29).

Json.net has an internal method Newtonsoft.Json.Utilities.MiscellaneousUtils.ValueEquals(object objA, object objB) which utilises IEquatable to determine whether or not to write a property when DefaultValueHandling is specified.

The bulk of this fix occurs in Values{T1,T2}, Values{T1,T2, T3}, Values{T1,T2, T3, T4}, which contain method overrides specific to the issue.

I have also added an additional test that was previously removed.

### DIFF
--- a/Source/Schema.NET.Tool/ViewModels/Property.cs
+++ b/Source/Schema.NET.Tool/ViewModels/Property.cs
@@ -35,7 +35,7 @@ namespace Schema.NET.Tool.ViewModels
                 }
                 else
                 {
-                    return $"Values<{adjustedTypesString}>?";
+                    return $"Values<{adjustedTypesString}>";
                 }
             }
         }
--- a/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
+++ b/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
@@ -27,7 +27,7 @@ namespace Schema.NET
             this.Value2 = default;
             this.Value3 = default;
             this.Value4 = default;
-            this.HasValue1 = value.Count > 0 ? true : false;
+            this.HasValue1 = value.Count > 0;
             this.HasValue2 = false;
             this.HasValue3 = false;
             this.HasValue4 = false;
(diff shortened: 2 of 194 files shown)


shervinw:
Also to note, running the generator causes the build to fail due to incompatible types, manually removed the following extract from [Thing.cs](Source/Schema.NET/core/Thing.cs) to match master. 

```C#
/// <summary>
/// Gets the context for the object, specifying that it comes from schema.org.
/// </summary>
[DataMember(Name = "@context", Order = 0)]
public override string Context => "http://schema.org";
```
The generator also modifies a bunch of properties to be of type `Values<T>` rather than `OneOrMany<T>`.

RehanSaeed:
Thanks @shervinw for contributing! Have not forgotten about this PR. I'll try to take a look early next week.

nickevansuk:
This looks exciting @shervinw! @RehanSaeed what do you think?

RehanSaeed:
I now have time to take a look at this. I've just upgraded the project to use .NET Core 3.0 and also publish NuGet packages to the GitHub package registry.

Bravo! A simple change that seems to solve all issues. Not sure why I missed it.

Added you as a contributor on the main page as thanks.

RehanSaeed:
Will release a major version update shortly.

nickevansuk:
Great news guys! Well done @shervinw !

@RehanSaeed can I check that this lib will still be compatible with .NET Core 1 and 2? As our library built on top of it needs to be backwards compatible too...

RehanSaeed:
Yes it will. We still target netstandard1.1 and netstandard2.0. I only use .NET Core 3 SDK to build, test and pack the package. It also gives us HTML test results.

shervinw:
Awesome thanks

Thanks
Shervin
________________________________

PR #97, opened 2019-11-10
Fix DateTime deserialisation to respect timezones

nickevansuk:
Without `DateParseHandling = DateParseHandling.DateTimeOffset`, the timezone gets truncated during deserialisation. Need to port some tests in to go along with this but thought I'd get this in while I was working on it.

Anyone very welcome to add a quick encode/decode test to show this working!

### DIFF
--- a/Source/Schema.NET/Thing.Partial.cs
+++ b/Source/Schema.NET/Thing.Partial.cs
@@ -19,7 +19,8 @@ namespace Schema.NET
                 new StringEnumConverter()
             },
             DefaultValueHandling = DefaultValueHandling.Ignore,
-            NullValueHandling = NullValueHandling.Ignore
+            NullValueHandling = NullValueHandling.Ignore,
+            DateParseHandling = DateParseHandling.DateTimeOffset
         };
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


Source/Schema.NET/Thing.Partial.cs
```diff
@@ -34,8 +35,35 @@ public partial class Thing : JsonLdObject
...
+            Converters = new List<JsonConverter>()
+            {
+                new StringEnumConverter()
+            },
+            NullValueHandling = NullValueHandling.Ignore,
+            DefaultValueHandling = DefaultValueHandling.Ignore,
+            DateParseHandling = DateParseHandling.DateTimeOffset,
+            // The ASP.NET MVC framework defaults to 32 (so 32 levels deep in the JSON structure)
+            // to prevent stack overflow caused by malicious complex JSON requests.
+            MaxDepth = 32
```
RehanSaeed: Apart from MaxDepth this is the same as the one above. Is there a reason to keep them separate?

Source/Schema.NET/Thing.Partial.cs
```diff
@@ -34,8 +35,35 @@ public partial class Thing : JsonLdObject
...
+        };
+
+        /// <summary>
+        /// Returns a strongly typed model of the JSON-LD representation provided.
+        /// </summary>
+        /// <typeparam name="T">Type of schema.org object to deserialize (can use Thing for any)</typeparam>
+        /// <param name="str">JSON string</param>
+        /// <returns>Strongly typed schema.org model</returns>
+        public static T Deserialize<T>(string str) => JsonConvert.DeserializeObject<T>(str, DeserializerSettings);
```
RehanSaeed: This is a reasonable addition. If we do this, then we should add an overload that takes a stream. Also, we should document this addition in the main README.md file.

RehanSaeed:
Do you have time to carry this forward or do you need help?

Source/Schema.NET/Thing.Partial.cs
nickevansuk: No reason to keep them separate, MaxDepth could be applied to both (though not really necessary for serialisation)

Source/Schema.NET/Thing.Partial.cs
nickevansuk: Great idea!

nickevansuk:
Hey @RehanSaeed sorry for being slow to move this forward. I'm shortly going on leave until the new year so my attention has been focussed elsewhere.

Help very much appreciated :)

Source/Schema.NET/Thing.Partial.cs
nickevansuk: Also on this line - we should also add a constraint of `where T : Schema.NET.JsonLdObject` (see https://github.com/openactive/OpenActive.NET/blob/master/OpenActive.NET/OpenActiveSerializer.cs#L125 for example)

RehanSaeed:
No problem, I'm in a similar situation. Will see if I can update your PR with these changes if I have time or maybe you will get there first.

RehanSaeed:
@Turnerj Have you seen this PR. The `MaxDepth` is a nice security addition and I'm not sure about the date handling.

Turnerj:
`MaxDepth` is interesting, hadn't thought about that yet. While we still use `Newtonsoft.Json`, it probably is a good idea to include a `MaxDepth` value as the default is no max depth. For `System.Text.Json`, the default is 64 `MaxDepth`.

Regarding the `DateTimeOffset` changes, that should be fixed from our changes to the `ValuesJsonConverter` for reading values as covered by these tests:

https://github.com/RehanSaeed/Schema.NET/blob/d9c7c7b6220655f0bd142f27b714d59a6cb4d432/Tests/Schema.NET.Test/ValuesJsonConverterTest.cs#L194-L216

Turnerj:
We also now have [a custom `SchemaSerializer` class](https://github.com/RehanSaeed/Schema.NET/blob/master/Source/Schema.NET/SchemaSerializer.cs) that is our middle-man to whatever serializer technology is underneath, it wouldn't be a big change to add `MaxDepth` to our deserialization options.

PR #100, opened 2019-11-20
Switch to System.Text.Json

### LINKED ISSUE #96: Switch from Newtonsoft.Json to System.Text.Json
System.Text.Json is now built into .NET and is the preferred library for JSON manipulation.

Turnerj:
Attempts to resolve #96 , switching from JSON.Net to System.Text.Json.

One problem with the migration is it requires dropping .NET Standard 1.1 support. I've done that in this PR already however if you're wanting to maintain that, it might be possible with falling back to JSON.Net but likely will require a ton of `#if` declarations in every class.

Current status: Down from 27000 errors to 0 - just testing is left.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,8 @@ jobs:
           dotnet-version: "5.0.x"
       - name: "Install .NET Core SDK"
         uses: actions/setup-dotnet@v1.8.2
+        with:
+          include-prerelease: true
       - name: "Dotnet Tool Restore"
         run: dotnet tool restore
         shell: pwsh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,8 +43,6 @@ jobs:
           dotnet-version: "5.0.x"
       - name: "Install .NET Core SDK"
         uses: actions/setup-dotnet@v1.8.2
-        with:
-          include-prerelease: true
       - name: "Dotnet Tool Restore"
         run: dotnet tool restore
         shell: pwsh
--- a/Schema.NET.sln
+++ b/Schema.NET.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.31515.178
+# Visual Studio Version 17
+VisualStudioVersion = 17.0.31612.314
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Source", "Source", "{719809C2-A551-4C4A-9EFD-B10FB5E35BC0}"
 	ProjectSection(SolutionItems) = preProject
--- a/Schema.NET.sln
+++ b/Schema.NET.sln
@@ -94,6 +94,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Schema.NET.Pending", "Sourc
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Common", "Common", "{AF228B8A-7290-4E26-8506-934C290C6AE3}"
 	ProjectSection(SolutionItems) = preProject
+		Source\Common\Constants.cs = Source\Common\Constants.cs
 		Source\Common\ContactType.cs = Source\Common\ContactType.cs
 		Source\Common\ContextJsonConverter.cs = Source\Common\ContextJsonConverter.cs
 		Source\Common\DateTimeHelper.cs = Source\Common\DateTimeHelper.cs
--- a/Schema.NET.sln
+++ b/Schema.NET.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.0.31423.177
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.31515.178
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Source", "Source", "{719809C2-A551-4C4A-9EFD-B10FB5E35BC0}"
 	ProjectSection(SolutionItems) = preProject
@@ -106,6 +106,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Common", "Common", "{AF228B
 		Source\Common\JsonLdObject.cs = Source\Common\JsonLdObject.cs
 		Source\Common\OneOrMany{T}.cs = Source\Common\OneOrMany{T}.cs
 		Source\Common\PropertyValueSpecification.Partial.cs = Source\Common\PropertyValueSpecification.Partial.cs
+		Source\Common\SchemaEnumJsonConverter{T}.cs = Source\Common\SchemaEnumJsonConverter{T}.cs
 		Source\Common\SchemaSerializer.cs = Source\Common\SchemaSerializer.cs
 		Source\Common\Thing.Partial.cs = Source\Common\Thing.Partial.cs
 		Source\Common\TimeSpanToISO8601DurationValuesJsonConverter.cs = Source\Common\TimeSpanToISO8601DurationValuesJsonConverter.cs
--- a/Source/Common/ContextJsonConverter.cs
+++ b/Source/Common/ContextJsonConverter.cs
@@ -41,7 +41,7 @@ namespace Schema.NET
 
                 if (document.RootElement.TryGetProperty("name", out var nameElement))
                 {
-                    name = nameElement.GetString() ?? "http://schema.org";
+                    name = nameElement.GetString() ?? Constants.HttpsSchemaOrgUrl;
                 }
 
                 if (document.RootElement.TryGetProperty("@language", out var languageElement))
@@ -74,10 +74,8 @@ namespace Schema.NET
                 }
             }
 
-#pragma warning disable CA1062 // Validate arguments of public methods
             context.Name = name;
             context.Language = language;
-#pragma warning restore CA1062 // Validate arguments of public methods
             return context;
         }
 
--- a/Source/Common/ContextJsonConverter.cs
+++ b/Source/Common/ContextJsonConverter.cs
@@ -1,8 +1,8 @@
 namespace Schema.NET
 {
     using System;
-    using Newtonsoft.Json;
-    using Newtonsoft.Json.Linq;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
 
     /// <summary>
     /// Converts a <see cref="JsonLdContext"/> object to and from JSON.
@@ -11,78 +11,65 @@ namespace Schema.NET
     public class ContextJsonConverter : JsonConverter<JsonLdContext>
     {
         /// <inheritdoc />
-        public override JsonLdContext ReadJson(JsonReader reader, Type objectType, JsonLdContext? existingValue, bool hasExistingValue, JsonSerializer serializer)
+        public override JsonLdContext Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
 #if NET6_0_OR_GREATER
-            ArgumentNullException.ThrowIfNull(reader);
-            ArgumentNullException.ThrowIfNull(objectType);
-            if (hasExistingValue)
-            {
-                ArgumentNullException.ThrowIfNull(existingValue);
-            }
-
-            ArgumentNullException.ThrowIfNull(serializer);
+            ArgumentNullException.ThrowIfNull(typeToConvert);
+            ArgumentNullException.ThrowIfNull(options);
 #else
-            if (reader is null)
+            if (typeToConvert is null)
             {
-                throw new ArgumentNullException(nameof(reader));
+                throw new ArgumentNullException(nameof(typeToConvert));
             }
 
-            if (objectType is null)
+            if (options is null)
             {
-                throw new ArgumentNullException(nameof(objectType));
-            }
-
-            if (hasExistingValue && existingValue is null)
-            {
-                throw new ArgumentNullException(nameof(existingValue));
-            }
-
-            if (serializer is null)
-            {
-                throw new ArgumentNullException(nameof(serializer));
+                throw new ArgumentNullException(nameof(options));
             }
 #endif
+            var context = new JsonLdContext();
 
-            var context = hasExistingValue ? existingValue! : new JsonLdContext();
-
-            string? name;
-            string? language;
-            if (reader.TokenType == JsonToken.String)
+            string? name = null;
+            string? language = null;
+            if (reader.TokenType == JsonTokenType.String)
             {
-                name = (string?)reader.Value;
-                language = null;
+                name = reader.GetString();
             }
-            else if (reader.TokenType == JsonToken.StartObject)
+            else if (reader.TokenType == JsonTokenType.StartObject)
             {
-                var o = JObject.Load(reader);
+                var document = JsonDocument.ParseValue(ref reader);
 
-                var nameProperty = o.Property("name", StringComparison.OrdinalIgnoreCase);
-                name = nameProperty?.Value?.ToString() ?? "https://schema.org";
+                if (document.RootElement.TryGetProperty("name", out var nameElement))
+                {
+                    name = nameElement.GetString() ?? "http://schema.org";
+                }
 
-                var languageProperty = o.Property("@language", StringComparison.OrdinalIgnoreCase);
-                language = languageProperty?.Value?.ToString();
+                if (document.RootElement.TryGetProperty("@language", out var languageElement))
+                {
+                    language = languageElement.GetString();
+                }
             }
             else
             {
-                var a = JArray.Load(reader);
+                var array = JsonDocument.ParseValue(ref reader).RootElement.EnumerateArray();
 
-                name = language = null;
-                foreach (var entry in a)
+                foreach (var entry in array)
                 {
-                    if (entry.Type == JTokenType.String)
+                    if (entry.ValueKind == JsonValueKind.String)
                     {
-                        name ??= (string?)entry;
+                        name ??= entry.GetString();
                     }
-                    else
+                    else if (entry.ValueKind == JsonValueKind.Object)
                     {
-                        var o = (JObject)entry;
+                        if (entry.TryGetProperty("name", out var nameElement))
+                        {
+                            name ??= nameElement.GetString() ?? "http://schema.org";
+                        }
 
-                        var nameProperty = o.Property("name", StringComparison.OrdinalIgnoreCase);
-                        name ??= nameProperty?.Value?.ToString() ?? "https://schema.org";
-
-                        var languageProperty = o.Property("@language", StringComparison.OrdinalIgnoreCase);
-                        language ??= languageProperty?.Value?.ToString();
+                        if (entry.TryGetProperty("@language", out var languageElement))
+                        {
+                            language ??= languageElement.GetString();
+                        }
                     }
                 }
             }
@@ -95,12 +82,12 @@ namespace Schema.NET
         }
 
         /// <inheritdoc />
-        public override void WriteJson(JsonWriter writer, JsonLdContext? value, JsonSerializer serializer)
+        public override void Write(Utf8JsonWriter writer, JsonLdContext value, JsonSerializerOptions options)
         {
 #if NET6_0_OR_GREATER
             ArgumentNullException.ThrowIfNull(writer);
             ArgumentNullException.ThrowIfNull(value);
-            ArgumentNullException.ThrowIfNull(serializer);
+            ArgumentNullException.ThrowIfNull(options);
 #else
             if (writer is null)
             {
@@ -112,23 +99,23 @@ namespace Schema.NET
                 throw new ArgumentNullException(nameof(value));
             }
 
-            if (serializer is null)
+            if (options is null)
             {
-                throw new ArgumentNullException(nameof(serializer));
+                throw new ArgumentNullException(nameof(options));
             }
 #endif
 
             if (string.IsNullOrWhiteSpace(value.Language))
             {
-                writer.WriteValue(value.Name);
+                writer.WriteStringValue(value.Name);
             }
             else
             {
                 writer.WriteStartObject();
                 writer.WritePropertyName("name");
-                writer.WriteValue(value.Name);
+                writer.WriteStringValue(value.Name);
                 writer.WritePropertyName("@language");
-                writer.WriteValue(value.Language);
+                writer.WriteStringValue(value.Language);
                 writer.WriteEndObject();
             }
         }
--- a/Source/Common/ContextJsonConverter.cs
+++ b/Source/Common/ContextJsonConverter.cs
@@ -63,7 +63,7 @@ namespace Schema.NET
                     {
                         if (entry.TryGetProperty("name", out var nameElement))
                         {
-                            name ??= nameElement.GetString() ?? "http://schema.org";
+                            name ??= nameElement.GetString() ?? Constants.HttpsSchemaOrgUrl;
                         }
 
                         if (entry.TryGetProperty("@language", out var languageElement))
--- a/Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
+++ b/Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
@@ -30,11 +30,6 @@ namespace Schema.NET
                 throw new ArgumentNullException(nameof(writer));
             }
 
-            if (value is null)
-            {
-                throw new ArgumentNullException(nameof(value));
-            }
-
             if (options is null)
             {
                 throw new ArgumentNullException(nameof(options));
--- a/Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
+++ b/Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
@@ -2,7 +2,8 @@ namespace Schema.NET
 {
     using System;
     using System.Globalization;
-    using Newtonsoft.Json;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
 
     /// <summary>
     /// Converts an <see cref="IValues"/> object to JSON. If the <see cref="IValues"/> contains a
@@ -17,8 +18,8 @@ namespace Schema.NET
         /// </summary>
         /// <param name="writer">The JSON writer.</param>
         /// <param name="value">The value to write.</param>
-        /// <param name="serializer">The JSON serializer.</param>
-        public override void WriteObject(JsonWriter writer, object? value, JsonSerializer serializer)
+        /// <param name="options">The JSON serializer options.</param>
+        public override void WriteObject(Utf8JsonWriter writer, object? value, JsonSerializerOptions options)
         {
 #if NET6_0_OR_GREATER
             ArgumentNullException.ThrowIfNull(writer);
@@ -29,19 +30,24 @@ namespace Schema.NET
                 throw new ArgumentNullException(nameof(writer));
             }
 
-            if (serializer is null)
+            if (value is null)
             {
-                throw new ArgumentNullException(nameof(serializer));
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (options is null)
+            {
+                throw new ArgumentNullException(nameof(options));
             }
 #endif
 
             if (value is DateTime dateTimeType)
             {
-                writer.WriteValue(dateTimeType.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture));
+                writer.WriteStringValue(dateTimeType.ToString("yyyy-MM-dd", CultureInfo.InvariantCulture));
             }
             else
             {
-                base.WriteObject(writer, value, serializer);
+                base.WriteObject(writer, value, options);
             }
         }
     }
--- a/Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
+++ b/Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
@@ -23,7 +23,7 @@ namespace Schema.NET
         {
 #if NET6_0_OR_GREATER
             ArgumentNullException.ThrowIfNull(writer);
-            ArgumentNullException.ThrowIfNull(serializer);
+            ArgumentNullException.ThrowIfNull(options);
 #else
             if (writer is null)
             {
--- a/Source/Common/EnumHelper.cs
+++ b/Source/Common/EnumHelper.cs
@@ -9,9 +9,6 @@ namespace Schema.NET
     /// </summary>
     internal static class EnumHelper
     {
-        private const int HttpSchemaOrgLength = 18; // equivalent to "http://schema.org/".Length
-        private const int HttpsSchemaOrgLength = 19; // equivalent to "https://schema.org/".Length
-
         /// <summary>
         /// Converts the string representation of the name or numeric value of one or more
         /// enumerated constants to an equivalent enumerated object.
@@ -67,13 +64,13 @@ namespace Schema.NET
             if (value is not null && value.StartsWith(Constants.HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
             {
 #pragma warning disable IDE0057 // Use range operator. Need to multi-target.
-                enumString = value.Substring(HttpSchemaOrgLength);
+                enumString = value.Substring(Constants.HttpSchemaOrgUrl.Length);
 #pragma warning restore IDE0057 // Use range operator. Need to multi-target.
             }
             else if (value is not null && value.StartsWith(Constants.HttpsSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
             {
 #pragma warning disable IDE0057 // Use range operator. Need to multi-target.
-                enumString = value.Substring(HttpsSchemaOrgLength);
+                enumString = value.Substring(Constants.HttpsSchemaOrgUrl.Length);
 #pragma warning restore IDE0057 // Use range operator. Need to multi-target.
             }
             else
--- a/Source/Common/EnumHelper.cs
+++ b/Source/Common/EnumHelper.cs
@@ -1,6 +1,7 @@
 namespace Schema.NET
 {
     using System;
+    using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
 
     /// <summary>
@@ -8,6 +9,11 @@ namespace Schema.NET
     /// </summary>
     internal static class EnumHelper
     {
+        private const string HttpSchemaOrgUrl = "http://schema.org/";
+        private const int HttpSchemaOrgLength = 18; // equivalent to "http://schema.org/".Length
+        private const string HttpsSchemaOrgUrl = "https://schema.org/";
+        private const int HttpsSchemaOrgLength = 19; // equivalent to "https://schema.org/".Length
+
         /// <summary>
         /// Converts the string representation of the name or numeric value of one or more
         /// enumerated constants to an equivalent enumerated object.
@@ -42,6 +48,50 @@ namespace Schema.NET
             return Enum.TryParse(enumType, value, out result);
 #pragma warning restore IDE0022 // Use expression body for methods
 #endif
+        }
+
+        /// <summary>
+        /// Converts the Schema URI representation of the enum type to an equivalent enumerated object.
+        /// </summary>
+        /// <param name="enumType">The enum type to use for parsing.</param>
+        /// <param name="value">The string representation of the name or numeric value of one or more enumerated constants.</param>
+        /// <param name="result">When this method returns true, an object containing an enumeration constant representing the parsed value.</param>
+        /// <returns><see langword="true"/> if the conversion succeeded; <see langword="false"/> otherwise.</returns>
+        public static bool TryParseEnumFromSchemaUri(
+            Type enumType,
+#if NETCOREAPP3_1_OR_GREATER
+            [NotNullWhen(true)]
+#endif
+            string? value,
+            out object? result)
+        {
+            string? enumString;
+            if (value is not null && value.StartsWith(HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
+            {
+#pragma warning disable IDE0057 // Use range operator. Need to multi-target.
+                enumString = value.Substring(HttpSchemaOrgLength);
+#pragma warning restore IDE0057 // Use range operator. Need to multi-target.
+            }
+            else if (value is not null && value.StartsWith(HttpsSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
+            {
+#pragma warning disable IDE0057 // Use range operator. Need to multi-target.
+                enumString = value.Substring(HttpsSchemaOrgLength);
+#pragma warning restore IDE0057 // Use range operator. Need to multi-target.
+            }
+            else
+            {
+                enumString = value;
+            }
+
+            if (TryParse(enumType, enumString, out result))
+            {
+                return true;
+            }
+            else
+            {
+                Debug.WriteLine($"Unable to parse enumeration of type {enumType.FullName} with value {enumString}.");
+                return false;
+            }
         }
     }
 }
--- a/Source/Common/EnumHelper.cs
+++ b/Source/Common/EnumHelper.cs
@@ -9,9 +9,7 @@ namespace Schema.NET
     /// </summary>
     internal static class EnumHelper
     {
-        private const string HttpSchemaOrgUrl = "http://schema.org/";
         private const int HttpSchemaOrgLength = 18; // equivalent to "http://schema.org/".Length
-        private const string HttpsSchemaOrgUrl = "https://schema.org/";
         private const int HttpsSchemaOrgLength = 19; // equivalent to "https://schema.org/".Length
 
         /// <summary>
@@ -66,13 +64,13 @@ namespace Schema.NET
             out object? result)
         {
             string? enumString;
-            if (value is not null && value.StartsWith(HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
+            if (value is not null && value.StartsWith(Constants.HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
             {
 #pragma warning disable IDE0057 // Use range operator. Need to multi-target.
                 enumString = value.Substring(HttpSchemaOrgLength);
 #pragma warning restore IDE0057 // Use range operator. Need to multi-target.
             }
-            else if (value is not null && value.StartsWith(HttpsSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
+            else if (value is not null && value.StartsWith(Constants.HttpsSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
             {
 #pragma warning disable IDE0057 // Use range operator. Need to multi-target.
                 enumString = value.Substring(HttpsSchemaOrgLength);
--- a/Source/Common/EnumHelper.cs
+++ b/Source/Common/EnumHelper.cs
@@ -64,13 +64,13 @@ namespace Schema.NET
             if (value is not null && value.StartsWith(Constants.HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
             {
 #pragma warning disable IDE0057 // Use range operator. Need to multi-target.
-                enumString = value.Substring(Constants.HttpSchemaOrgUrl.Length);
+                enumString = value.Substring(Constants.HttpSchemaOrgUrl.Length + 1);
 #pragma warning restore IDE0057 // Use range operator. Need to multi-target.
             }
             else if (value is not null && value.StartsWith(Constants.HttpsSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
             {
 #pragma warning disable IDE0057 // Use range operator. Need to multi-target.
-                enumString = value.Substring(Constants.HttpsSchemaOrgUrl.Length);
+                enumString = value.Substring(Constants.HttpsSchemaOrgUrl.Length + 1);
 #pragma warning restore IDE0057 // Use range operator. Need to multi-target.
             }
             else
--- a/Source/Common/SchemaEnumJsonConverter{T}.cs
+++ b/Source/Common/SchemaEnumJsonConverter{T}.cs
@@ -41,6 +41,11 @@ namespace Schema.NET
         /// <returns>The enumeration value.</returns>
         public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
+            if (typeToConvert is null)
+            {
+                throw new ArgumentNullException(nameof(typeToConvert));
+            }
+
             var valueString = reader.GetString();
             if (EnumHelper.TryParseEnumFromSchemaUri(typeToConvert, valueString, out var result))
             {
@@ -56,6 +61,14 @@ namespace Schema.NET
         /// <param name="writer">The JSON writer.</param>
         /// <param name="value">The enumeration value.</param>
         /// <param name="options">The JSON serializer options.</param>
-        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options) => writer.WriteStringValue(this.valueNameMap[value]);
+        public override void Write(Utf8JsonWriter writer, T value, JsonSerializerOptions options)
+        {
+            if (writer is null)
+            {
+                throw new ArgumentNullException(nameof(writer));
+            }
+
+            writer.WriteStringValue(this.valueNameMap[value]);
+        }
     }
 }
--- a/Source/Common/SchemaEnumJsonConverter{T}.cs
+++ b/Source/Common/SchemaEnumJsonConverter{T}.cs
@@ -1,0 +1,61 @@
+namespace Schema.NET
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+    using System.Runtime.Serialization;
+    using System.Text.Json;
+    using System.Text.Json.Serialization;
+
+    /// <summary>
+    /// Converts a Schema enumeration to and from JSON.
+    /// </summary>
+    /// <typeparam name="T">The enum type to convert</typeparam>
+    public class SchemaEnumJsonConverter<T> : JsonConverter<T>
+        where T : struct, Enum
+    {
+        private readonly Dictionary<T, string> valueNameMap = new();
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SchemaEnumJsonConverter{T}"/> class.
+        /// </summary>
+        public SchemaEnumJsonConverter()
+        {
+            var enumType = typeof(T);
+            var values = Enum.GetValues(enumType);
+
+            foreach (var value in values)
+            {
+                var enumMember = enumType.GetMember(value!.ToString()!)[0];
+                var enumMemberAttribute = enumMember.GetCustomAttribute<EnumMemberAttribute>(false);
+                this.valueNameMap[(T)value] = enumMemberAttribute!.Value!;
+            }
+        }
+
+        /// <summary>
+        /// Reads the JSON representation of the enumeration.
+        /// </summary>
+        /// <param name="reader">The <see cref="Utf8JsonReader"/> to read from.</param>
+        /// <param name="typeToConvert">Type of the enumeration.</param>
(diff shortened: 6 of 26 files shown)


Turnerj:
It doesn't look like some of the functionality that JSON.Net has exists in System.Text.Json, specifically the handling of default values. Currently JSON.Net is configured to ignore default values [however you can't do that with System.Text.Json](https://stackoverflow.com/q/58226377/1676444).

Assuming I can get everything else to work, I think this still might be a deal breaker.

Turnerj:
A few more opinionated changes in the `Thing` class too with the `ToString` methods.

The short of it is:
- [HTML Escaping is enabled by default in `JsonSerializer`](https://stackoverflow.com/a/58504215/1676444) so I've obsoleted the `ToHtmlEscapedString` method. It can be disabled by using a custom encoder however with it being enabled by default, might be a good time to drop this method - up to you!
- The static `JsonSerializerOptions` need to use a static constructor due to the access level of the `Converters` property (its read only).

Turnerj:
Currently through getting the tests working, there is a big blocker with: dotnet/corefx#38841

I could manually change every `DataContract` and `DataMember` reference to use the new `JsonPropertyName` property but not sure if that is a good idea.

RehanSaeed:
This is good stuff. I wondered if there would be road blocks like null value handling problems.

With regards to HTML escaping by default. There are times when you might not want to do that (you're not working in a web context), so we need to provide both methods.

Regarding https://github.com/dotnet/corefx/issues/38841, it may be that we need to wait for .NET Core 3.1 before this PR can be accepted.

Can you not set the order with System.Text.Json at all? I wouldn't mind a complete switch to the new attributes if required. We don't support any other serialization like XML, so there is no benefit to using `DataMember`.

Turnerj:
As far as I've seen in the source code, no it doesn't seem possible to set the order with the new property. I have a feeling they probably won't have that functionality though as according to the JSON spec, an object is an unordered list of name/value pairs.

As far as I can tell, the serializer will put inherited properties at the end so it is at least still predictable, just that all the tests that do a string comparison will need to have their properties moved around.

I've updated the `Thing.ToString()` methods so the HTML escaping and non-escaping paths are working as intended again.

I'm happy to wait for 3.1 to see if `DataMember` change has landed though can move to `JsonPropertyName` if you want. 🙂

RehanSaeed:
Happy to move to `JsonPropertyName`, it's just a find and replace to get there.

The ordering is a bigger problem as you have to mess with all of the tests which is a bigger job and much harder to review the PR. If there are no plans to implement ordering in 3.1, then perhaps we should go ahead with it, otherwise wait for 3.1. Thoughts?

Turnerj:
I really don't see them bringing across ordering to `JsonPropertyName` simply because JSON has no dependency on ordering of object keys normally. At a quick count, there is maybe 36 JSON strings in the tests so it is too much to modify.

It does suck needing to change the tests, less because of the work to do it, more because it is far nicer having the code meet the tests than updating both (code and tests) to meet each other.

I'll look at getting this done soon, even if it is just so we can see if there are any other deal breakers with switching to `System.Text.Json`.

Turnerj:
Note for self: Need to update `Schema.NET.Tool` - it has a number of string references to `DataMember` and `Newtonsoft` that would need updating.

RehanSaeed:
Instead of comparing exact string representations in our tests, perhaps we should compare JSON object types from System.Text.Json?

Turnerj:
That could work! So with Newtonsoft that was `JObject` and in `System.Text.Json` that is `JsonDocument` - it would likely make the code less fragile. It also would have a built in comparer so besides writing up the model, it should fit easily with the test assertions.

The only exception to this is still specifically checking how the `ToString` extensions on `Thing` work. They specifically replace the inner `@context` properties from the JSON and this wouldn't match the normal deserialization process.

RehanSaeed:
Yes, I hadn't thought of that. I guess string comparisons are still important. The System.Text.Json API must still output a deterministic order for all it's properties, so I guess it might be best to bite the bullet and fix the tests to match.

Turnerj:
Hitting a hard error at the moment around `System.Text.Json` either not respecting or ignoring the `ValuesJsonConverter`. It does perform the `CanConvert` check however after that, it does some additional check where it figures it is impossible to turn `String` to `Array`.

It doesn't even call the read method on the `ValuesJsonConverter` which makes this extra weird. If it did call it, I would figure I had a bug in that code. With it not calling it, I'm gonna have to dive into `System.Text.Json` to work out why it even does this.

Fun fact with this though, using `System.Text.Json` 4.6.0 (latest stable) has a harder to understand exception, just saying it couldn't cast `String` to `OneOrMany<String>`. Updating to the latest 4.7.0 preview was able to give me this detail.

```
Message: 
    System.Text.Json.JsonException : The JSON value could not be converted to Schema.NET.OneOrMany`1[System.String]. Path: $.name | LineNumber: 0 | BytePositionInLine: 77.
    ---- System.InvalidOperationException : The requested operation requires an element of type 'Array', but the target element has type 'String'.
  Stack Trace: 
    ThrowHelper.ReThrowWithPath(ReadStack& readStack, Utf8JsonReader& reader, Exception ex)
    JsonSerializer.ReadCore(JsonSerializerOptions options, Utf8JsonReader& reader, ReadStack& readStack)
    JsonSerializer.ReadCore(Type returnType, JsonSerializerOptions options, Utf8JsonReader& reader)
    JsonSerializer.Deserialize(String json, Type returnType, JsonSerializerOptions options)
    JsonSerializer.Deserialize[TValue](String json, JsonSerializerOptions options)
    MusicEventTest.Deserializing_MusicEventJsonLd_ReturnsMusicEvent() line 97
    ----- Inner Stack Trace -----
    JsonDocument.CheckExpectedType(JsonTokenType expected, JsonTokenType actual)
    JsonDocument.GetArrayLength(Int32 index)
    JsonElement.GetArrayLength()
    ValuesJsonConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options) line 50
    JsonPropertyInfoNotNullableContravariant`4.OnRead(ReadStack& state, Utf8JsonReader& reader)
    JsonPropertyInfo.Read(JsonTokenType tokenType, ReadStack& state, Utf8JsonReader& reader)
    JsonSerializer.ReadCore(JsonSerializerOptions options, Utf8JsonReader& reader, ReadStack& readStack)
```

Turnerj:
Another note to self: There looks to be a more efficient way to cast raw JSON to value types than using `int.TryParse` etc. This would be used in `ValuesJsonConverter` where many of the direct `TryParse` methods are used directly.

See: https://stackoverflow.com/a/57334833/1676444

Turnerj:
Had to move to a preview version of `System.Text.Json` as the code was hitting an unavoidable error. It seems like `System.Text.Json` 4.6.0 doesn't understand `OneOrMany` properly and while it uses the `CanConvert` on `ValuesJsonConverter`, it never actually used the converter itself. Anyway, updating to the latest preview avoids that issue.

A few tests are now passing with no change (eg. `MusicEventTest`) though I'm not sure some of these tests are valid. `MusicEventTest` for example deserializes and serializes again a string which wouldn't actually test whether the string JSON is an exact match. It might be worth looking through some of the tests to make sure they are testing the right things. The vast majority seem to be doing that (in at least, most of them are still failing) but some passed where they likely shouldn't have.

Turnerj:
Two curly issues I've needed to work around due to how `System.Text.Json` works though they are a result of the same problem. The serializer doesn't seem to recognize object types properly, instead falling back to whatever interface on a collection or specific generic parameter provided is.

Case in point, the `ToString` method on `Thing` passed `this` which made the serializer think it was only wanting to serialize the properties of `Thing` (rather than the actual type that inherits it). That was a brain buster to solve but simply passing the type directly resolves it.

Similarly when serializing a `OneOrMany` property, they normally point to the interface (eg. `OneOrMany<ICreativeWork>`) which made the serializer only try to serialize that interface and the problems that arise from that (not using the right JSON property names). The work around for this was looping the enumerable directly rather than passing it off.

Ultimately though we are still stuck with default values being serialized. The "IgnoreNullValues" requires classes but because the property types are structs, that never works so the serialized JSON is full of `null` values. This sets back actually updating the tests - we might have to wait for dotnet/corefx#38878 before continuing much more on this PR.

Turnerj:
I threw together a basic "Thing" converter which mitigates the need for `System.Text.Json` to support ignoring default values. I've put it in a Gist as I'm not sure we want to go that direction or just wait for real support: https://gist.github.com/Turnerj/11bb3832e4d4a40464768311de3d46e8

Turnerj:
I've fixed issue #104 for `System.Text.Json` though the current master with JSON.Net still has the issue described.

RehanSaeed:
Do you think we're close with this or should I start to think about fixing it with JSON.NET?

Turnerj:
For the most part, yeah we are close. The biggest issue is still the lack of ignoring default values - [I have a work around](https://gist.github.com/Turnerj/11bb3832e4d4a40464768311de3d46e8) which basically takes over all serialization of all schema entities, just needs the `JsonConverter` attribute on every class (not difficult). I'm not sure if we want to go this route or wait till real support for ignoring default values exists - it could be a while waiting.

Until we know the path for ignoring default values (whether to use my workaround or wait till real support), there isn't much point altering the tests. Doing so currently will still lead to test failures because of a lot of `null` properties being serialized and that would take a while to write out in all the JSON string assertions.

If you're happy with me using my workaround, I can continue but otherwise, it is a waiting game for dotnet/corefx#38878 to be done.

nickevansuk:
Just throwing my 2cents in here:

Really exciting to see this working, and great how much progress we’ve been able to make and contribute back to corefx issues at the same time to help steer System.Text.Json.

I help maintain a library (OpenActive.NET) on top of Schema.NET - expecting this change - so great to see its progressing.

My quick thought was given that this library is in use in production across a large number of systems, I’m not sure we need to push ourselves to be white-knuckle-bleeding-edge here? Perhaps sticking with the tried and tested JSON.NET while System.Text.Json settles into mainstream use would be best - especially now we’ve done the hard work to identify which bits of System.Text.Json we need to see improvements in before migrating?

Although we gain performance in theory using System.Text.Json (though perhaps not so much with the workaround still in place?) it doesn’t seem like there’s any obvious functional benefit here and we’re increasing complexity by being an early adopter?

As I said just my 2cents, is really great to be able to give the MS team feedback on where we’ve found issues with the new functionality, just not sure if we need to go the whole way just yet

Turnerj:
Those are fair points you've raised @nickevansuk - I too am working on a pretty substantial project where Schema.NET plays a vital part so I definitely want to make sure it is stable, fast and memory efficient.

I don't imagine the workaround for default values performing any slower than having that functionality built-in (both do caching certain reflected property lookups) but it would increase the overall code maintenance required for the library. On the other hand, once working support for ignoring default values lands, it would make it redundant so it likely wouldn't be around long-term.

Keeping with JSON.Net for now also does allow the library to keep .NET Standard 1.1 support (that was one of the first things that had to go for `System.Text.Json` support). Personally I'm indifferent about supporting .NET Standard 1.1, I am myself using .NET Core 3 but even then, there still is support for newer versions of .NET Framework etc.

What might be good in the meantime is actually beefing up the tests (I found some while doing this PR that don't seem to be adequately testing what I think they should be), adding some basic benchmarking (both to optimise the current code and to measure the performance improvement when going to `System.Text.Json`) and any other tweaking (a few little quality-of-life improvements when de-serializing certain values that don't specify a type, potentially auto-generated `IEquatable<T>` for each type, other useful things like merging schema objects).

If @RehanSaeed is happy for those types of things, I'm happy to focus on those things and let this PR sit till `System.Text.Json` has "ignore default value" support.

Turnerj:
If we still want to go ahead with `System.Text.Json` with my workaround, I would definitely be doing significant testing and validation to make sure it all plays nice - no point switching to a better dependency if everything is a hot mess of buggy code!

RehanSaeed:
Happy to wait for .NET 5 if it makes it easier. In that case we need to fix https://github.com/RehanSaeed/Schema.NET/issues/104 but I don't have much time in the short term.

Also as an aside if we want to add System.Text.Json, we should consider adding an option to serialize directly to UTF8 which is a bit faster. I've done this with some other code I've been working on but not given it much thought with Schema.NET.

RehanSaeed:
See docs about serializing to and from UTF8

https://docs.microsoft.com/en-gb/dotnet/standard/serialization/system-text-json-how-to

Turnerj:
I've done a preliminary benchmark run on `System.Text.Json` and while it isn't stable (one of the benchmarks failed - might be an internal `System.Text.Json` thing) and includes default values (so more strings it has to write), it performs about as good as my overhaul of the `ValuesJsonConverter` in #109.

|      Method |       Runtime |     Mean |   Error |  StdDev |      Min |      Max |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |-------------- |---------:|--------:|--------:|---------:|---------:|--------:|------:|------:|----------:|
|   Serialize |    .NET 4.7.2 | 297.3 us | 2.93 us | 2.74 us | 288.4 us | 299.3 us | 37.5977 |     - |     - |  119669 B |
| Deserialize |    .NET 4.7.2 |       NA |      NA |      NA |       NA |       NA |       - |     - |     - |         - |
|   Serialize | .NET Core 3.0 | 153.6 us | 2.58 us | 2.41 us | 149.1 us | 157.9 us | 15.1367 |     - |     - |   48024 B |
| Deserialize | .NET Core 3.0 | 319.1 us | 4.36 us | 3.87 us | 307.5 us | 324.7 us | 33.6914 |     - |     - |  106906 B |

I did have another thought regarding improving serialization/deserialization performance by having Schema.NET.Tool generate custom converters for each type. Technically we have all the pieces we need in the tool with each property and the ancestor properties etc plus we know all the types ahead of time.

Unless `System.Text.Json` is emitting a custom converter for each type, it has to be using some form of reflection for properties. Even if those are cached, I can't see it being faster than a custom converter for writing.

I think this is something worth pursuing and could actually make the need for ignoring default values redundant. Also, we would have control of the property order so the tests could stay as they are.

Turnerj:
With https://github.com/dotnet/runtime/pull/36322 - it looks like we will be closer to moving this forward and not needing to worry about default values. Will keep my eye out if there are any beta versions of the System.Text.Json library out with this new change so we can keep progressing on this.

Turnerj:
I've reworked this PR back on top of the changes we've had in master and have it using the latest System.Text.Json version (v5.0.0 Preview 4). Unfortunately there seems to be some major issues in this version of System.Text.Json as I get back a "Common Language Runtime detected an invalid program".

Hopefully the next preview version will have that fixed.

AraHaan:
@Turnerj is it possible that somewhere the enum is used as an nullable? I did that with my code and that is how it happened, look at everything that uses the enum, if anything adds an ``?`` to end of the enum type that is used in the json try removing the ``?`` and try again.

Turnerj:
Hey @AraHaan - Yeah, we use nullable types for pretty much everything internally. All the schema objects are generated based on the definitions by schema.org so we really don't have a say whether they can or can't be nullable.

In our tests though, we have specific ones that are meant to try out various types. For example, we have this test here for testing string values through our custom converter:

https://github.com/RehanSaeed/Schema.NET/blob/77413e5e5594130960fec1fe7f7198c985ed937f/Tests/Schema.NET.Test/ValuesJsonConverterTest.cs#L469-L474

This test [also failed](https://ci.appveyor.com/project/RehanSaeed/schema-net/builds/33064765/job/62lxs509wsfre175#L2698) with the same error so I'm not sure it is limited to enum or even value types specifically. I'll cross-post this to the other issue.

Turnerj:
Raised a new issue (dotnet/runtime#40878) about the serialization problem we are having with System.Text.Json

Turnerj:
I have some clarification on the issue - System.Text.Json is currently programmed to [intentionally _not_ box value types](https://github.com/dotnet/runtimelab/blob/JsonCodeGen/src/libraries/System.Text.Json/docs/SerializerProgrammingModel.md#performance) for custom converters.

We have two solutions currently with a potentially third if [this PR](https://github.com/dotnet/runtime/pull/40914) gets merged into the runtime which at this stage, I can't tell if that would be likely for .NET 5 let alone merged at all (going back to the intentional-ness of not supporting boxing already). I am happy to wait and see over the next week or two if that PR gets any movement - if not, we should probably consider one of the options below.

**Solution One: Change how our custom converter works**
If we have custom converters for each struct we have (`OneOrMany`, `Values<T1,T2>`, `Values<T1,T2,T3>` etc), we can more directly tell S.T.J what is going on and it works around the issue. [A super awesome person created a PR in my minified test case repository of Schema.NET.](https://github.com/Turnerj/SchemaNet_SystemTextJson/pull/1)

The converter would look like so:
```csharp
    public class OneOrManyJsonConverter<T> : JsonConverter<OneOrMany<T>>
    {
        [return: MaybeNull]
        public override OneOrMany<T> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            return (OneOrMany<T>)ValuesJsonConverter.Instance.Read(ref reader, typeToConvert, options);
        }

        public override void Write(Utf8JsonWriter writer, OneOrMany<T> value, JsonSerializerOptions options)
        {
            ValuesJsonConverter.Instance.Write(writer, value, options);
        }
    }

    public class ValuesJsonConverter<T1, T2> : JsonConverter<Values<T1, T2>>
    {
        [return: MaybeNull]
        public override Values<T1, T2> Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
        {
            return (Values<T1, T2>)ValuesJsonConverter.Instance.Read(ref reader, typeToConvert, options);
        }

        public override void Write(Utf8JsonWriter writer, Values<T1, T2> value, JsonSerializerOptions options)
        {
            ValuesJsonConverter.Instance.Write(writer, value, options);
        }
    }
```

I don't think this is a bad solution - it would require a little bit of work with changing the schema generation tool to output these altered converters on properties. We'd also need to change how the `TimeSpanToISO8601DurationValuesJsonConverter` and `
DateTimeToIso8601DateValuesJsonConverter` converters work in that case too.

**Solution Two: Change our `OneOrMany` and `Value<>` structs to classes**
This one is fairly simple - classes aren't boxed so avoids the problem entirely. In theory it wouldn't be any worse for the GC than the internal boxing that S.T.J would have done anyway.

I would do this as a separate PR and measure the effects with Newtonsoft - maybe even worth publishing a version with just that change if we consider that change breaking.

What's your thoughts @RehanSaeed ? Personally I like the simplicity to Solution Two though Solution One _may_ allow for an optimization or two in terms of having the generic parameters available in our converter.

RehanSaeed:
I haven't kept up with the changes in System.Text.Json since it came out.

Having a few JSON converters doesn't sound too onerous. How much of a performance difference do you think each method will have? At the end of the day, that's the main difference I think (or is there something else I'm missing?). Perhaps we should benchmark it?

Turnerj:
Good news - with dotnet/runtime#40914 being merged in (and back ported to .NET 5 via dotnet/runtime#41366), we shouldn't need to change anything on our side. 🎉

It isn't in Preview 8 that has just been released but I'm guessing it will be in one of the two RC versions before the full release. Hopefully then it means we can potentially get the last few bits of this done (tests need updating as the property order has changed) and finally merge the PR. 🙂

RehanSaeed:
That's awesome. Can't wait!

hansmbakker:
.Net 5.0 RC1 is released: https://devblogs.microsoft.com/dotnet/announcing-net-5-0-rc-1/

Turnerj:
The 5.0 RC1 does fix our big issue though it seems like we have some new ones:
- Default/null values for OneOrMany/Values etc are still serialized regardless of the ignore condition options
- The "Type" property is serialized twice (once as "@type" and once as "Type") due to how we override values. Attempting to ignore the overridden value causes neither to write.
- Seems to be a few cases where number values don't perfectly serialize back and forth (eg. 4000.0 != 4000), I think this is S.T.J trying to optimise something but haven't determined what we'd need to do.

With the first issue, it seems that `default(OneOrMany<T>)` (for example) isn't actually seen as a default value by S.T.J so when it passes the value to our `Write` method on `ValuesJsonConverter`, it writes `null`. We can't _not_ write `null` here as attempting to causes invalid JSON. I'll be raising an issue against the runtime about this.

With the "Type" property, while there is potential for workarounds, we'd need to take over writing properties from S.T.J which we wouldn't want to do. I'll be raising an issue against the runtime about this.

The only known issue that still needs addressing on our end relates to the property order as S.T.J serialization has property order dependent on inheritance so that "@context" is actually the last property, not the first. Given that JSON makes no distinction about property order, this is just us needing to update our tests.

Turnerj:
Just after writing my other comment, it does seem that the default value issue can be avoided by changing `OneOrMany` and `Values` to `class`... It really seems like `struct` is making our life harder. I'll still raise an issue about it to the runtime because it does seem like broken behaviour but it might be worth considering changing from `struct` to `class`.

Turnerj:
OK - I realised my test was bad re. `@type` and `Type` - that only occurred in my test because I was missing one final `[JsonPropertyName("@type")]` attribute.

So the only issue is just the default value handling that I need to raise with the .NET team.

Turnerj:
Good news is that [a fix was written and merged into the runtime](https://github.com/dotnet/runtime/pull/42319). Bad news is that [it doesn't seem that it will make .NET 5 (or perhaps even 5.1)](https://github.com/dotnet/runtime/pull/42406) as there is a potential work around (changing from `struct` to `class`).

RehanSaeed:
Lets wait for 5 and then review where we're at.

devsko:
>(changing from struct to class)

I would not call this a workaround

Turnerj:
> 
> 
> > (changing from struct to class)
> 
> I would not call this a workaround

I'm just using the language from [the linked PR](https://github.com/dotnet/runtime/pull/42406) description:

> Edge case scenario that has _workaround_

Later [in a comment when the PR was closed](https://github.com/dotnet/runtime/pull/42406#issuecomment-694522741):

> We don't expect many apps to hit this and can provide guidance on how to _workaround_ it in 5.0.

I can only assume the workaround is changing from `struct` to `class` because I have no idea what else the workaround would be.

Turnerj:
Not seeing any new issues with S.T.J 6.0.0-preview.2 _however_ our next roadblock seems to be property ordering. Currently it defaults to reflection order which follows the order of properties in the class, then of its parents. This means that properties like `@context` appear at the end.

We knew about this before but I considered it more of a testing problem than a user problem. The more I think about it, the more I realise that we should keep our property order as we've had it. It isn't a huge problem but it might be just annoying enough for some people plus makes debugging in general a pain.

See dotnet/runtime#728 for keeping track of serialization ordering being implemented.

If it doesn't land in the full release of S.T.J v6, that will be very annoying (for me anyway) as it means this PR has been delayed by more than 2 years!

RehanSaeed:
If they fixed it so that base class properties appear at the beginning, rather than at the end, then that would be ok for us right? Or do you think that we should always have a particular ordering to the properties? Apart from the annoyance at having to fix the tests, the only thing I can think of is consumers of our library may also have a reliance on ordering (most likely for tests again though).

Turnerj:
> If they fixed it so that base class properties appear at the beginning, rather than at the end, then that would be ok for us right?

It would be close with the exception of the `@type` property which we override in child classes. That property would appear incorrectly.

> Or do you think that we should always have a particular ordering to the properties?

From years ago dealing with a lot of JS, one of the things I remember was that JS objects didn't guarantee a key order. With JSON, APIs did _seem_ to return properties in a logical key order (in a similar way as your suggestion in the other PR about the ordering of constructor arguments).

On a parsing level, it generally wouldn't matter. You could (in theory) optimize for cases where you know the next parameter in sequence however from our POV, we don't deal with that level of parsing not to mention that the properties are optional and our values are mixed.

For a debugging POV, those core properties we have ordered first (context, type, id, title, name, description) are pretty useful so them being first in the JSON does make debugging easier. Because the ordering is affected by the hierarchy, it may seem to consumers that our ordering is arbitrary and make finding a particular property harder (of course though, they probably can press `Ctrl+F` in whatever program they are looking at the JSON in and find what they are looking for).

> Apart from the annoyance at having to fix the tests, the only thing I can think of is consumers of our library may also have a reliance on ordering (most likely for tests again though).

Yeah. The only other thought I had was caching - someone a hash of the JSON string we generate won't match any more.

-----

This is a bit of a side issue but I'm thinking with our tests to change a few things. I know we test a bunch of the models however it isn't exactly the models we need to test. I do think it is important we do a deserialize-serialize-deserialize test to prove there is no loss of data but that might be better against a model we directly control anyway rather than one we generate.

It is a big move but I'm thinking we remove the entire "core" folder and all the tests in it. They aren't really testing anything we weren't already testing except they were our generated types. The properties of the types are all `Values` and `OneOrMany` which we test pretty thoroughly so really, it is just testing our equality and hashcode methods which we don't actually call.

Removing the tests ironically would also remove our own issues with the property order too.

RehanSaeed:
> It is a big move but I'm thinking we remove the entire "core" folder and all the tests in it. They aren't really testing anything we weren't already testing except they were our generated types. The properties of the types are all Values and OneOrMany which we test pretty thoroughly so really, it is just testing our equality and hashcode methods which we don't actually call.

People have found these useful as they show examples of how to use the library. I've closed several issues just because I could point them at the common types that we have tests for. So I think they have more value than just being tests.

Turnerj:
> People have found these useful as they show examples of how to use the library. I've closed several issues just because I could point them at the common types that we have tests for. So I think they have more value than just being tests.

Hmmmm, that's a good point. I wonder though if we then could change a bit about what the tests are asserting. Rather than assert every property is what value we expect, we could make it a bit lighter? That way we have less code we need to maintain?

RehanSaeed:
Perhaps we should consider a snapshot testing tool like https://github.com/VerifyTests/Verify. I've had great success when using them with HTML in JavaScript.

AraHaan:
Since this was delayed a ton should this project enable C# 10, use the .NET 6 rc.1 SDK for the github actions (I made an action that always installs the daily quality .NET SDK builds) and then ~~use~~ I mean abuse file-scoped namespaces and global usings?

Also note with global usings I found an .NET SDK bug too: https://github.com/dotnet/sdk/issues/19446 but it should hopefully not get delayed for me long.

Turnerj:
> Since this was delayed a ton should this project enable C# 10, use the .NET 6 rc.1 SDK for the github actions (I made an action that always installs the daily quality .NET SDK builds which is https://github.com/Elskom/setup-latest-dotnet) and then use I mean abuse file-scoped namespaces and global usings?

Personally I'd really like to get this merged in before going down that path. It has been hard avoiding conflicts etc when rebasing so minimising potentially big changes across lots of files is ideal.

> As for testing what is the status so far?

The main issue is that some of our tests compare JSON strings and the property order is different between JSON.NET and System.Text.Json. It does look like the next preview might add `JsonPropertyOrder` which should resolve that allowing me to get all the tests passing.

_Technically_ I could manually change all the property orders manually in the tests but I really don't want to.

Anyway, once the property orders are correct I can finally re-run the benchmarks to see the performance difference. May try and add dotnet/runtime#31274 support to avoid more unnecessary allocations too (we currently turn `JsonElement` to a string and re-parse it again just to deserialize it to an object properly).

Sometime after this is all good and merged in, then I can potentially look at #310 and see if we can avoid reflection.

Turnerj:
I'm getting only 20 failing tests locally which relate to two things:
- Decimal places for numbers
- Enum values

The first is a little annoying - for example:
```
                                     ↓ (pos 390)
    Expected: ···ition","price":47200,"priceCurrency":"USD","priceValidUntil":···
    Actual:   ···ition","price":47200.0,"priceCurrency":"USD","priceValidUntil···
                                     ↑ (pos 390)
```

The problem here isn't as simple as adding a decimal place to our expected data. The expected result is actually from serializing a C# object and the actual is from deserializing then serializing again. So something is strange in this process. Worst case, we write a converter for decimals.

The second one is due to the `EnumMember` attribute not being used by System.Text.Json. We use `JsonStringEnumConverter` but will need to check if there is another attribute we can use or whether we write our own converter for it.

AraHaan:
Isn't there an option of striping the decimal spot if the decimal numbers are all 0?

Turnerj:
Not that I've seen - just stuff for whether numbers can be strings. If you do find a way to ignore unneeded decimals, I'm keen to hear it!

Turnerj:
So close... apparently CI is getting some build errors I don't get locally with analyzer rules (not sure why). Besides that, I know there is a minor issue with the `JsonLdContext` tests that I need to look into too.

The enum issue is solved with a customer converter (though I'd drop that in a heartbeat once S.T.J supports `EnumMember` - dotnet/runtime#31081). The decimal issue is "solved" but I'm not a huge fan - basically its going `ToString("G0")` on the number and writing it raw in our `ValuesJsonConverter`.

AraHaan:
So how close is this now to merge?

Turnerj:
The commit I've pushed just now should have all tests green! Now it really is fine tuning and having @RehanSaeed review the changes.

Currently with .NET 5, this PR is ~13% faster at serializing while allocating ~14% less. When deserializing, it is ~36% faster while allocating 40% less.

## Main Branch (Book Benchmark)

|      Method |                  Job |              Runtime |     Mean |   Error |  StdDev |      Min |      Max |   Gen 0 | Allocated |
|------------ |--------------------- |--------------------- |---------:|--------:|--------:|---------:|---------:|--------:|----------:|
|   Serialize |             .NET 5.0 |             .NET 5.0 | 131.2 us | 1.98 us | 3.19 us | 126.9 us | 139.0 us | 14.4043 |     45 KB |
| Deserialize |             .NET 5.0 |             .NET 5.0 | 138.7 us | 2.77 us | 3.08 us | 134.7 us | 146.4 us | 25.1465 |     77 KB |
|   Serialize | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 152.5 us | 1.76 us | 1.56 us | 150.4 us | 155.3 us | 14.4043 |     45 KB |
| Deserialize | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 178.8 us | 3.16 us | 2.80 us | 173.3 us | 180.9 us | 26.1230 |     80 KB |

## This PR (Book Benchmark)

|      Method |                  Job |              Runtime |      Mean |    Error |   StdDev |       Min |       Max |   Gen 0 | Allocated |
|------------ |--------------------- |--------------------- |----------:|---------:|---------:|----------:|----------:|--------:|----------:|
|   Serialize |             .NET 5.0 |             .NET 5.0 | 115.14 us | 2.183 us | 2.144 us | 113.10 us | 120.15 us | 12.5732 |     39 KB |
| Deserialize |             .NET 5.0 |             .NET 5.0 |  88.66 us | 1.170 us | 0.977 us |  87.14 us |  89.91 us | 15.0146 |     46 KB |
|   Serialize | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 144.76 us | 2.868 us | 3.522 us | 140.10 us | 151.45 us | 12.4512 |     39 KB |
| Deserialize | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 153.50 us | 1.983 us | 1.855 us | 148.26 us | 156.08 us | 15.8691 |     49 KB |

Turnerj:
Unfortunately dotnet/runtime#31274 didn't land in Preview 7 so I couldn't low allocations further.

Source/Common/ContextJsonConverter.cs
```diff
@@ -11,46 +11,34 @@ namespace Schema.NET
     public class ContextJsonConverter : JsonConverter<JsonLdContext>
     {
         /// <inheritdoc />
-        public override JsonLdContext ReadJson(JsonReader reader, Type objectType, JsonLdContext? existingValue, bool hasExistingValue, JsonSerializer serializer)
+        public override JsonLdContext Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
         {
-            if (reader is null)
+            if (typeToConvert is null)
```
RehanSaeed, line 16: We should null check all parameters.

Source/Common/ContextJsonConverter.cs
```diff
@@ -11,46 +11,34 @@ namespace Schema.NET
...
             else
             {
-                var o = JObject.Load(reader);
+                var o = JsonDocument.ParseValue(ref reader);
 
-                var nameProperty = o.Property("name", StringComparison.OrdinalIgnoreCase);
-                name = nameProperty?.Value?.ToString() ?? "https://schema.org";
+                if (o.RootElement.TryGetProperty("name", out var nameElement))
+                {
+                    name = nameElement.GetString() ?? "http://schema.org";
```
RehanSaeed, line 35: We should move `"http://schema.org"` into a constant somethere. Don't we have one already?

Source/Common/ContextJsonConverter.cs
```diff
@@ -11,46 +11,34 @@ namespace Schema.NET
...
 
-                var languageProperty = o.Property("@language", StringComparison.OrdinalIgnoreCase);
-                language = languageProperty?.Value?.ToString();
+                if (o.RootElement.TryGetProperty("@language", out var languageElement))
+                {
+                    language = languageElement.GetString();
+                }
             }
 
 #pragma warning disable CA1062 // Validate arguments of public methods
```
RehanSaeed, line 44: We can now remove this warning suppression. The weird parameters of JSON.NET are gone.

Source/Common/DateTimeToIso8601DateValuesJsonConverter.cs
```diff
@@ -17,26 +18,31 @@ public class DateTimeToIso8601DateValuesJsonConverter : ValuesJsonConverter
...
-                throw new ArgumentNullException(nameof(serializer));
+                throw new ArgumentNullException(nameof(value));
+            }
+
+            if (options is null)
+            {
+                throw new ArgumentNullException(nameof(options));
             }
 
             if (value is DateTime dateTimeType)
```
RehanSaeed, line 39: Now that we have the `DateOnly` type coming, we'll need to rethink this strategy.

Source/Common/EnumHelper.cs
```diff
@@ -1,13 +1,19 @@
...
     using System.Diagnostics.CodeAnalysis;
 
     /// <summary>
     /// Helper for parsing strings into Enum values.
     /// </summary>
     internal static class EnumHelper
     {
+        private const string HttpSchemaOrgUrl = "http://schema.org/";
+        private const int HttpSchemaOrgLength = 18; // equivalent to "http://schema.org/".Length
+        private const string HttpsSchemaOrgUrl = "https://schema.org/";
```
RehanSaeed, line 14: Here's that constant we were looking for.

Source/Common/SchemaEnumJsonConverter{T}.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <summary>
+        /// Reads the JSON representation of the enumeration.
+        /// </summary>
+        /// <param name="reader">The <see cref="Utf8JsonReader"/> to read from.</param>
+        /// <param name="typeToConvert">Type of the enumeration.</param>
+        /// <param name="options">The serializer options.</param>
+        /// <returns>The enumeration value.</returns>
+        public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            if (typeToConvert is null)
```
RehanSaeed, line 44: Should we null check all arguments?

Source/Common/SchemaSerializer.cs
```diff
@@ -13,50 +13,43 @@ public static class SchemaSerializer
...
         /// <summary>
         /// Deserializes the JSON to the specified type.
         /// </summary>
         /// <typeparam name="T">Deserialization target type</typeparam>
         /// <param name="value">JSON to deserialize</param>
         /// <returns>An instance of <typeparamref name="T"/> deserialized from JSON</returns>
-        public static T? DeserializeObject<T>(string value) =>
-            JsonConvert.DeserializeObject<T>(value, DeserializeSettings);
+        public static T? DeserializeObject<T>(string value)
+            => JsonSerializer.Deserialize<T>(value, DefaultSerializationSettings);
```
RehanSaeed, line 52: In the latest editorconfig changes in VS 16.11 you can decide to have `=>` or `&&` or `||` at the beginning/end of the line. I went with the end, so I think this would create a warning in later versions of VS.

Source/Common/TimeSpanToISO8601DurationValuesJsonConverter.cs
```diff
@@ -30,18 +31,18 @@ public override void WriteObject(JsonWriter writer, object? value, JsonSerialize
                 throw new ArgumentNullException(nameof(value));
```
RehanSaeed, line 31: Should we be throwing here? `value` is nullable.

Source/Common/ValuesJsonConverter.cs
```diff
@@ -6,104 +6,81 @@ namespace Schema.NET
...
+        /// <param name="reader">The <see cref="Utf8JsonReader"/> to read from.</param>
+        /// <param name="typeToConvert">Type of the object.</param>
+        /// <param name="options">The serializer options.</param>
         /// <returns>The object value.</returns>
-        public override object? ReadJson(
-            JsonReader reader,
-            Type objectType,
-            object? existingValue,
-            JsonSerializer serializer)
+        public override IValues? Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
```
RehanSaeed, line 43: We should check `reader` for `null` here.

Source/Common/ValuesJsonConverter.cs
```diff
@@ -128,28 +105,27 @@ public override void WriteJson(JsonWriter writer, object? value, JsonSerializer
                 throw new ArgumentNullException(nameof(value));
             }
 
-            if (serializer is null)
+            if (options is null)
```
RehanSaeed, line 108: Should we be checking `value` for `null` here.

Source/Common/ValuesJsonConverter.cs
```diff
@@ -383,11 +354,19 @@ private static bool TryProcessTokenAsType(JsonReader reader, Type targetType, ou
                     result = localResult;
                 }
             }
-            else if (tokenType is JsonToken.Integer or JsonToken.Float)
+            else if (tokenType == JsonTokenType.Number)
+            {
+                if (targetType == typeof(short) || targetType == typeof(int) || targetType == typeof(long) || targetType == typeof(float) || targetType == typeof(double) || targetType == typeof(decimal))
```
RehanSaeed, line 359: Should we be checking for all number types here? Including unsighned types?

Source/Schema.NET/Schema.NET.csproj
```diff
@@ -14,7 +14,7 @@
   </PropertyGroup>
 
   <ItemGroup Label="Package References">
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="System.Text.Json" Version="6.0.0-preview.7.21377.19" />
```
RehanSaeed, line 17: We have to wait for .NET 6 to be released before merging right?

RehanSaeed (commented):
Made a few small comments.

Schema.NET.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 17
-VisualStudioVersion = 17.0.31423.177
+# Visual Studio Version 16
```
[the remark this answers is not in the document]
Turnerj: Nope, not using VS 2022 yet - totally didn't realise the solution header updated in the PR. Must have been when I added a new file to the "Common" code solution folder.

Source/Common/ContextJsonConverter.cs
Turnerj: Not sure if we do but that's a good idea. Also, should it be `https://schema.org`?

Source/Common/ContextJsonConverter.cs
Turnerj: We don't actually use the `options` parameter - would it still be worth adding checks?

Source/Common/EnumHelper.cs
Turnerj: Ahhhhh, yeah. Should we move this out somewhere else so it doesn't sit in the `EnumHelper` if we are referencing it in the `JsonLdContext` class?

Source/Common/SchemaEnumJsonConverter{T}.cs
Turnerj: It's another case like the other one you noted - we don't actually use `options` so it didn't seem necessary. The `reader` is a `struct` so we're fine in that regard.

Source/Common/TimeSpanToISO8601DurationValuesJsonConverter.cs
Turnerj: That's a good point - this is wrong.

Source/Schema.NET/Schema.NET.csproj
Turnerj: I mean, this is actually compiling with the .NET 5 SDK. I'd be fine having it merged prior to the release of .NET 6 but it does limit any other releases we might want to do between now and then.

If that isn't much of an issue, I think perhaps we merge and release a preview version of Schema.NET with S.T.J to see what issues etc crop up that we've missed?

Source/Common/ValuesJsonConverter.cs
Turnerj: Reader can't be null, it is a `struct`. Not sure if we should be checking then if its `default` then but my understanding is that we can trust it to be OK.

Source/Common/ValuesJsonConverter.cs
Turnerj: Maybe? We don't have any `Values` types on our models that have unsigned values but people who extend Schema.NET might.

Source/Common/ValuesJsonConverter.cs
Turnerj: Probably a bit lower down, yeah - we write a `null` if the `IValues` has a count of 0 but we don't actually check if its `null`.

Source/Schema.NET/Schema.NET.csproj
AraHaan: I think it's worth it to update github actions to the .NET 6 SDK.

Note: you can every easily do it with my version of setup-dotnet that is far simpler than setup-dotnet called setup-latest-dotnet 😂. Yes I made it for the reason of getting the daily quality builds for github actions based on a specific major and minor version feed name for .NET.

Source/Common/ContextJsonConverter.cs
RehanSaeed: In that case, I guess we can leave it.

Source/Common/ContextJsonConverter.cs
RehanSaeed: Yes, `https`. Good catch.

Source/Common/SchemaEnumJsonConverter{T}.cs
RehanSaeed: Agreed

Source/Common/SchemaSerializer.cs
RehanSaeed: I've been very frustrated with `.editorconfig` since its so buggy and has some undocumented settings. I wouldn't be surprised if you didn't see a warning.

Source/Schema.NET/Schema.NET.csproj
RehanSaeed: @AraHaan We only need to update the `global.json` file. I've setup everything (GitHub actions, Azure Pipelines, Appveyor) to get the SDK version from there.

Schema.NET.sln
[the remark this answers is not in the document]
Turnerj: I think I've addressed all your other feedback - just need to sort this out still.

Turnerj:
Slight performance boost and allocation reduction in deserialization now that dotnet/runtime#31274 is in effect via System.Text.Json RC1.

|      Method |                  Job |              Runtime |      Mean |    Error |   StdDev |       Min |       Max |   Gen 0 | Allocated |
|------------ |--------------------- |--------------------- |----------:|---------:|---------:|----------:|----------:|--------:|----------:|
|   Serialize |             .NET 5.0 |             .NET 5.0 | 119.30 us | 2.315 us | 3.604 us | 114.69 us | 127.45 us | 12.4512 |     39 KB |
| Deserialize |             .NET 5.0 |             .NET 5.0 |  84.93 us | 1.596 us | 1.493 us |  81.95 us |  87.26 us | 13.0615 |     40 KB |
|   Serialize | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 143.78 us | 2.381 us | 2.227 us | 139.98 us | 146.82 us | 12.4512 |     39 KB |
| Deserialize | .NET Framework 4.7.2 | .NET Framework 4.7.2 | 148.28 us | 0.915 us | 0.856 us | 146.89 us | 149.85 us | 13.9160 |     43 KB |

The CI does seem to be having some issues with RC1 so I'll need to investigate that.

erri120:
> The CI does seem to be having some issues with RC1 so I'll need to investigate that.

You need to enable preview support:

```yaml
- uses: actions/setup-dotnet@v1
  with:
    dotnet-version: '6.0.x'
    include-prerelease: true
```

(https://github.com/actions/setup-dotnet#usage)

Turnerj:
> 
> > The CI does seem to be having some issues with RC1 so I'll need to investigate that.
> 
> You need to enable preview support:
> 
> ```yaml
> - uses: actions/setup-dotnet@v1
>   with:
>     dotnet-version: '6.0.x'
>     include-prerelease: true
> ```
> 
> (https://github.com/actions/setup-dotnet#usage)

Yeah, it seems that I need to set `include-prerelease` at least. The version itself though should be coming from the `global.json` file - this way the CI installs the same SDK version that we run locally.

The root cause seems to relate to dotnet/sdk#21006, it sounds like it is just an SDK issue as the S.T.J source generator (the one we aren't using - yet (#310)) is still involved in the build process and it requires the later SDK.

Turnerj:
Hey @RehanSaeed - not sure how we should handle this. I think I'm doing the `global.json` correctly (it works locally) but seems to break the `setup-dotnet` action. If you have any ideas, you should be able to contribute directly to my branch.

We're like 1.5 months or so from the launch of .NET 6 - do we bother trying to get the RC1 version working before then?
I wouldn't mind getting this done prior to .NET 6 being out so we can release a beta version of it for testing but it is your call.  🙂

RehanSaeed:
There is also a preview VM in GitHub Actions and Azure Devops using VS 2022 Preview. I'm assuming we also need this and updating the global.json to enable preview versions.

AraHaan:
I actually released a much better action than setup-dotnet so that can be an option as well. Currently it defaults to using the .NET install scripts to download from the .NET 6 daily quality feeds (which is now 6.0.0-rtm that they are currently working on until November), After that they should have a 7.0 feed up and I will then commit onto it to default to the 7.0 daily quality feed.

See: https://github.com/Elskom/setup-latest-dotnet and https://github.com/Elskom/build-dotnet if you want more details.

Yes, I basically wanted to make github actions easier for more common things I was using it for in GitHub Actions, if you want build-dotnet to do other things as well on top of what it already supports feel free to file an issue for it 😄.

Turnerj:
> There is also a preview VM in GitHub Actions and Azure Devops using VS 2022 Preview. I'm assuming we also need this and updating the global.json to enable preview versions.

In theory it shouldn't matter what VM it is - the `setup-dotnet` action should be installing the SDK we say. From looking at the code behind it, it also didn't use the `allowPrerelease` setting that we can use in `global.json`.

> I actually released a much better action than setup-dotnet so that can be an option as well. Currently it defaults to using the .NET install scripts to download from the .NET 6 daily quality feeds (which is now 6.0.0-rtm that they are currently working on until November), After that they should have a 7.0 feed up and I will then commit onto it to default to the 7.0 daily quality feed.

While I can speak for @RehanSaeed , my thought isn't so much we need cutting edge or daily builds of the SDK - it just happens to be that _this_ PR needs the latest. Locally `dotnet build` etc uses our `global.json` file to know what version of the SDK it should be fixed to - I'd just want that to work correctly in all 3 CIs we use (GitHub Actions, Azure DevOps, AppVeyor). Having a fixed SDK locally and on the CI helps avoid weird build bugs that only occur in one environment but not the other - this is similar to the reason for running multiple different CIs as we have hit issues in some CIs but not others before.

The actual problem then seems to be a multi-part problem:
- GitHub Action "setup-dotnet" doesn't seem read `allowPrerelease`
- Even if it did, the `dotnet-install` script doesn't seem to handle `6.0.100-rc.1` as a version (the URL it tries to fetch the SDK from is invalid - [see this error](https://github.com/RehanSaeed/Schema.NET/pull/100/checks?check_run_id=3651135712#step:3:16)) - the value `6.0.100-rc.1` seems valid for `global.json` as my builds locally work fine with it so it is likely not an _invalid_ value.

RehanSaeed:
- I don't see an issue in `setup-dotnet` for `allowPrerelease` support.  In theory, I'd expect it to just work.
- `dotnet-install` has been buggy in the past. RC's don't come out often, I assume its broken due to that reason. We should raise a bug with them too.

Turnerj:
>     * I don't see an issue in `setup-dotnet` for `allowPrerelease` support.  In theory, I'd expect it to just work.

This is the only code in `setup-dotnet` that actually reads the `global.json` file: https://github.com/actions/setup-dotnet/blob/5a3fa01c67e60dba8f95e2878436c7151c4b5f01/src/setup-dotnet.ts#L51-L66

It only looks at `sdk.version` and `sdk.rollForward`.
 
>     * `dotnet-install` has been buggy in the past. RC's don't come out often, I assume its broken due to that reason. We should raise a bug with them too.

Found the repo for the source scripts: https://github.com/dotnet/install-scripts
Will look into raising an issue about getting RC versions to work

Turnerj:
Wooo! I'll rebase and fix any conflicts - if our tests pass, let's merge it in and do a new release.

Will do this in the next 10 minutes or so...

RehanSaeed:
Cool, after that I'd like to update the .editorconfig which will give us a code fix that can switch to file scoped namespaces.

Turnerj:
Took a bit longer to get sorted than I thought but it seems to all be good! Any final things you want to check before we merge @RehanSaeed ?

.github/workflows/build.yml
```diff
@@ -43,6 +43,8 @@ jobs:
           dotnet-version: "5.0.x"
       - name: "Install .NET Core SDK"
         uses: actions/setup-dotnet@v1.8.2
+        with:
```
RehanSaeed, line 46: We can drop this now.

Source/Common/ContextJsonConverter.cs
```diff
@@ -11,96 +11,81 @@ namespace Schema.NET
...
-                        var o = (JObject)entry;
-
-                        var nameProperty = o.Property("name", StringComparison.OrdinalIgnoreCase);
-                        name ??= nameProperty?.Value?.ToString() ?? "https://schema.org";
-
-                        var languageProperty = o.Property("@language", StringComparison.OrdinalIgnoreCase);
-                        language ??= languageProperty?.Value?.ToString();
+                        if (entry.TryGetProperty("name", out var nameElement))
+                        {
+                            name ??= nameElement.GetString() ?? "http://schema.org";
```
RehanSaeed, line 66: Should we use the constant here?

Source/Common/EnumHelper.cs
```diff
@@ -1,13 +1,17 @@
...
     using System;
+    using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
 
     /// <summary>
     /// Helper for parsing strings into Enum values.
     /// </summary>
     internal static class EnumHelper
     {
+        private const int HttpSchemaOrgLength = 18; // equivalent to "http://schema.org/".Length
```
RehanSaeed, line 12: Should we move to use the URL Constants with `.Length` or move these to the constants file?

RehanSaeed (commented):
LGTM minus a couple of very minor suggestions. Feel free to complete.

Source/Common/ContextJsonConverter.cs
Turnerj: Fixed! Missed this when merging the code.

Source/Common/EnumHelper.cs
Turnerj: Yep - using the constants with `.Length` now. Even if there is a penalty (there "shouldn't" be because our strings were constant), it should be extremely minor in the scheme of things.

.github/workflows/build.yml
Turnerj: Done!

Source/Common/EnumHelper.cs
Turnerj: There was a minor issue using the constant - there is a trailing slash here that the constant doesn't have. I decided to just `+ 1` the length value from the constant rather than duplicate the string etc. I did update the doc for the constants to mention it excludes the trailing slash.

AraHaan:
Yay, finally after a few years it was done. :smile:

PR #106, opened 2019-12-10
Fix link to Contributing guide

Turnerj:
Ironically I only noticed this from opening the other PR but basically, the link was broken for the contributing guide. The old link pointed to the repository root for the file but it actually exists in the `.github` folder.


RehanSaeed:
Thanks for that.

PR #111, opened 2019-12-14
Updated artifact path to match other build scripts

Turnerj:
Don't you just love that English has the same word with two different spellings?

Changing AppVeyor's use to match the Azure Pipelines use seems to be the path of least hassle.

Anyway, this should allow artifacts/artefacts to be picked up correctly as they currently aren't: https://ci.appveyor.com/project/RehanSaeed/schema-net/builds/29536122/job/3cckikl9vbq8vuk9#L196

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,10 +33,10 @@ test: off
 
 artifacts:
 - name: NuGet Packages
-  path: ./Artifacts/**/*.nupkg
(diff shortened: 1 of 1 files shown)


RehanSaeed:
Good spot.

I find my own self confused sometimes having grown up with the British English spellings but then developing code with the Americanised spellings. The age old question, do you go with 'correct' spelling or the American one?

PR #112, opened 2019-12-15
Add type equality checks and overrides

Turnerj:
All classes implement IEquatable and have GetHashCode and Equals overrides. Not sure if it is worth adding the operator overloads for equality too.


Source/Schema.NET.Tool/ViewModels/Class.cs
```diff
@@ -150,43 +150,72 @@ public override string ToString()
 
             stringBuilder.AppendIndentLine(4, "{");
 
-            if (string.Equals(this.Name, "Thing", StringComparison.OrdinalIgnoreCase))
```
RehanSaeed: Any reason for removing this property?

Source/Schema.NET/JsonLdObject.cs
```diff
@@ -43,6 +43,32 @@ public class JsonLdObject
...
+
+        /// <inheritdoc/>
+        public bool Equals(JsonLdObject other)
+        {
+            if (other is null)
+            {
+                return false;
+            }
+
+            if (ReferenceEquals(this, other))
```
RehanSaeed: In theory StyleCop would require `this.` before `ReferenceEquals`.

Tests/Schema.NET.Test/ThingTest.cs
```diff
@@ -28,5 +28,71 @@ public class ThingTest
             Assert.Equal(
                 this.thing.ToString(),
                 JsonConvert.DeserializeObject<Thing>(this.json, TestDefaults.DefaultJsonSerializerSettings).ToString());
+
```
RehanSaeed: Some derived type and mixed comparison tests would also be useful.

Source/Schema.NET.Tool/ViewModels/Class.cs
Turnerj: If it exists, it leads to an unbuildable project. `Context` already exists on `JsonLdObject` and is type `JsonLdContext`. The override declares the type as `string` (this is why it fails) and the schema URL. On top of that, `JsonLdContext` already defines the default value for the context with the HTTPS URL.

So even sorting the type issue, it seems a little redundant having `Thing` specifically set a value that is already defined by default.

Source/Schema.NET/JsonLdObject.cs
Turnerj: `ReferenceEquals` is a static method so I don't think we can actually use `this`. We could use `object` if we wanted though Visual Studio was telling me to remove that.

RehanSaeed:
Cool thanks.

We should bump the version. If we get some of your other changes in, it'd be a major version upgrade.

Turnerj:
Yeah - if we can my overhaul of the `ValueJsonConverter` merged (once we've sorted any other loose ends for it out), I think we'd be in a good place to version bump.

PR #113, opened 2019-12-16
Add testing support for .NET Framework

Turnerj:
When looking into some issues with #100, I realised that the unit tests aren't setup to test .NET Framework. For that PR at least, it shows a large number of issues with .NET Framework - I don't know if that is from the custom `ValuesJsonConverter` or from `System.Text.Json` itself.

In either case however, I realised it would be a good thing to have even outside of that PR (for example, testing #109). There is actually one failing test when testing .NET Framework on master: https://github.com/RehanSaeed/Schema.NET/blob/c815daf10ed59e9328992e0f729b2ee23529e62b/Tests/Schema.NET.Test/ValuesJsonConverterTest.cs#L238-L255

The test fails as the current `ValuesJsonConverter` on master returns `null` due to not finding a suitable type. This isn't a problem in .NET Core 3 with JSON.Net but is a problem in .NET Framework 4.7.2 - this issue however should already be resolved in #109 due to how it attempts to find a concrete type.

I'm not fully versed in the build process for Schema.NET so I don't know if it is going to be as easy as the two changes I've made. I guess we will see when the CI finishes...

If you are curious, the `StringExtensions` class is to work around the combination of multiple calls to these specific methods with those methods not being supported in .NET Framework and the linting requiring these calls in .NET Core. I didn't really want to wrap the individual calls in linting suppression statements.

### DIFF
--- a/Tests/Schema.NET.Test/Schema.NET.Test.csproj
+++ b/Tests/Schema.NET.Test/Schema.NET.Test.csproj
@@ -1,10 +1,13 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
     <TargetFrameworks>net472;netcoreapp3.1</TargetFrameworks>
     <CodeAnalysisRuleSet>../../MinimumRecommendedRulesWithStyleCop.ruleset</CodeAnalysisRuleSet>
     <LangVersion>latest</LangVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <RestoreAdditionalProjectSources>
+      https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
+    </RestoreAdditionalProjectSources>
   </PropertyGroup>
 
   <ItemGroup Label="Project References">
@@ -12,8 +15,6 @@
   </ItemGroup>
 
   <ItemGroup Label="Package References">
-    <!-- See https://github.com/Microsoft/dotnet/tree/master/releases/reference-assemblies -->
-    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.4.0" />
     <PackageReference Include="Moq" Version="4.13.1" />
     <PackageReference Include="xunit" Version="2.4.1" />
@@ -23,6 +24,14 @@
   <ItemGroup Label="Analyzer Package References">
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="All" Version="16.4.33" />
     <PackageReference Include="StyleCop.Analyzers" PrivateAssets="All" Version="1.1.118" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(OS)' != 'Windows_NT' AND '$(TargetFramework)' == 'net472' ">
+    <PackageReference
+        Include="Microsoft.TargetingPack.NETFramework.v4.7.2"
+        Version="1.0.0"
+        ExcludeAssets="All"
+        PrivateAssets="All" />
   </ItemGroup>
 
   <ItemGroup>
--- a/Tests/Schema.NET.Test/Schema.NET.Test.csproj
+++ b/Tests/Schema.NET.Test/Schema.NET.Test.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


Tests/Schema.NET.Test/Schema.NET.Test.csproj
```diff
@@ -23,6 +26,14 @@
     <PackageReference Include="StyleCop.Analyzers" PrivateAssets="All" Version="1.1.118" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(OS)' != 'Windows_NT' AND '$(TargetFramework)' == 'net472' ">
```
RehanSaeed: To add `net472` support with mono, you should only need this https://github.com/RehanSaeed/Schema.NET/blob/c815daf10ed59e9328992e0f729b2ee23529e62b/Source/Schema.NET/Schema.NET.csproj#L45 NuGet package.

RehanSaeed: I'm not actually sure what `Microsoft.TargetingPack.NETFramework.v4.7.2` is.

Tests/Schema.NET.Test/StringExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Schema.NET.Test
+{
+    using System;
+
+#pragma warning disable IDE0060 // Remove unused parameter
+#pragma warning disable CA1801 // Remove unused parameter
+
+    public static class StringExtensions
+    {
+        public static string Replace(this string target, string oldValue, string newValue, StringComparison stringComparison) => target.Replace(oldValue, newValue);
```
RehanSaeed: We could copy the code from corefx or at least throw when `StringComparison.Ordinal` is not used for both of these methods.

RehanSaeed: The warnings would then disappear.

Tests/Schema.NET.Test/Schema.NET.Test.csproj
RehanSaeed: Is it because of xUnit not supporting it? I think I remember something about this. This project used to target .NET 4.6.2 and I went through all kinds of gymnastics to get multi-tartgetting working.

Tests/Schema.NET.Test/StringExtensions.cs
Turnerj: This was just to make it happy with .NET Framework as it doesn't have those methods - even where we are usingthe `StringComparison`, it isn't needed, its really just to appease the linting. If we were to throw on not having `StringComparison`, the methods that rely on it would fail for .NET Framework.

Tests/Schema.NET.Test/Schema.NET.Test.csproj
Turnerj: Yeah so I had just that initially however while it compiled, the tests failed with some internal xUnit error - was just trying to work around it. Its the first time I heard of `Microsoft.TargetingPack.NETFramework.v4.7.2` too but apparently its a thing that sounds like it does what we want though still not working.

Most of my testing experience is with MSTest/VSTest/whatever-they-call-it-these-days so I don't know about any quirks specific to xUnit.

Turnerj: This is where I got the information about the targeting pack, from xUnit's own website: https://xunit.net/docs/getting-started/multi-target/non-windows

Turnerj: Turns out the result is no better using [`TargetingPack`](https://ci.appveyor.com/project/RehanSaeed/schema-net/builds/29558592/job/mslwxfmsmpm37un0#L232) vs [`ReferenceAssemblies`](https://ci.appveyor.com/project/RehanSaeed/schema-net/builds/29566908/job/3ao46h9mld2rtdfp#L232).

Turnerj: OK, got the tests running on .NET Framework. I've gone back to using `ReferenceAssemblies` because that is a whole lot neater and realised there was just one additional dependency needed. In 5c79c4e, I've added `Microsoft.TestPlatform.ObjectModel` and now the tests are running fine.

The failing test is known to fail for .NET Framework and should be fixed in #109.

RehanSaeed: Please note that I've noticed chocolatey and AppVeyor have been having weird issues over the last few days, so those builds have been a bit flaky.

I think that targetting pack stuff is now superceded by the new `Microsoft.NETFramework.ReferenceAssemblies` package. All references to it seem a bit older.

Turnerj:
With the .NET Framework tests running on each platform (with the CI failures coming from a specific test which should be fixed as part of #109), are you happy if we merge it in?

RehanSaeed:
I think it makes sense for us to get https://github.com/RehanSaeed/Schema.NET/pull/109 in first and come back to this.

RehanSaeed:
This should now build successfully, now that https://github.com/RehanSaeed/Schema.NET/pull/109 is in.

PR #114, opened 2019-12-18
Values Struct Equality Fix

Turnerj:
There was a subtle bug in the equality checking of the `Values` structs that didn't properly handle mixed types. While `Values2` did properly handle mixed types, it didn't properly handle only one of the types having data.

After a bit of studying the code, it really can be simplified down to comparing the internal `OneOrMany`. I do add one additional check prior to that as a shortcut for empty data.

### DIFF
--- a/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
+++ b/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
@@ -473,40 +473,15 @@ namespace Schema.NET
         /// </returns>
         public bool Equals(Values<T1, T2, T3, T4> other)
         {
-            if (other.HasValue1)
-            {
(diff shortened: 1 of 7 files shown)


Turnerj:
The test I've removed in cd68610 was flawed - the original `ValuesJsonConverter` never had a value deserializing into multiple types at once. The reason the test was passing up until now is that it played on a flaw in the equality method of the `Values` struct.

While I could have changed the expected data to only include a URI value (as that was the type of data that was deserialized), it would no longer fit with the spirit of the other tests in that file. It wouldn't really be any more of a test than the tests added specifically for `ValuesJsonConverter` or now for the various `Values` structs themselves.

RehanSaeed:
Thank you.

PR #117, opened 2019-12-19
Switch from .ruleset & GlobalSupressions.cs to .editorconfig files

RehanSaeed:
In the past we used three different ways to supress warnings:

- `.ruleset` files to supress warnings across the solution.
- `<NoWarn>` in a `.csproj` file to supress warnings in a project.
- `GlobalSupressions.cs` to supress warnings in a project or member.

In the later versions of Visual Studio you can now supress warnings using an `.editorconfig` file. This PR, switches completely to this new format which is much simpler and easier to reason about.

As a side effect, we now have slightly more warnings because we aren't turning them off at the solution level anymore, so I also fix those.

The `Schema.NET.Tool` project was moved to a `Tools` subdirectory because it's not something we ship. I wanted to do this in the past but Cake build also created a temporary `tools` folder but no longer does, since we started using the Cake global tool.


RehanSaeed:
@Turnerj A quick PR to keep the project fresh, learn and try new things and keep it using the latest tech.

RehanSaeed:
Yes, it's a very powerful tool now. You can override settings.

Issue #116, opened 2019-12-19
OneOrMany handling of Value types vs Reference types with default values


Turnerj:
I've been looking at further performance and allocation optimisations - one that I have been messing with is making `OneOrMany` only internally hold an array rather than an array and a single item.

While for the most part this is just an implementation change (no API changes), it did show a few things, that are strange to me, with how `OneOrMany` works.

[`OneOrMany` won't store a single null value (Count == 0)](https://github.com/RehanSaeed/Schema.NET/blob/master/Source/Schema.NET/OneOrMany%7BT%7D.cs#L96) but will allow multiple null values (Count > 1). This is strange behaviour though is backed up in the tests like in the `CarTest`. 

https://github.com/RehanSaeed/Schema.NET/blob/master/Tests/Schema.NET.Test/core/CarTest.cs#L20

I feel that it should be consistent - if any null values (or empty strings) are provided, they are ignored/not added internally to the `OneOrMany`. This is what the implicit conversion logic does but I think the constructor should be handling that instead. I would likely inline the implicit conversion logic to the constructor(s) instead.

The other thing I found is if `OneOrMany` is a value type, it always has a value. I understand this from the perspective that `int` etc can't be null but at the same time, it doesn't make sense to me why a default `OneOrMany<int>` actually has anything at all. I'm not sure if it is like this due to the previous implementation (by holding a single `item` value of `T` which would not be null for value types) or if it was intentional.

This is what I've currently built (builds upon changes in #109): https://gist.github.com/Turnerj/6d0400887912f454ef2be033c682ff4e#file-oneormany-cs

Really this issue is just to gauge your thoughts on such a change? Various tests would need updating as a number play on the logic that something like `OneOrMany<int>` should always have a value. (I can work to keep that logic but I figure while I'm doing this update, I might find out if it is even necessary)

I'm happy to do the changes - probably in #109 or otherwise after that has merged. Right now it looks like a 3% allocation improvement (in the Book benchmark) though I think there is probably more I can get out of it.

RehanSaeed:
> I feel that it should be consistent - if any null values (or empty strings) are provided, they are ignored/not added internally to the OneOrMany. This is what the implicit conversion logic does but I think the constructor should be handling that instead. I would likely inline the implicit conversion logic to the constructor(s) instead.

This sounds like a bug. The above sounds sensible.

[...]

PR #118, opened 2019-12-20
Refactoring OneOrMany

Turnerj:
Changes the internal logic of `OneOrMany` to be an array only (rather than an array or single item) as well as addresses a few minor bugs outlined in #116.

These changes also result in a ~3% allocation reducation in .NET Framework (though with no measurable reduction in .NET Core).

### DIFF
--- a/Source/Schema.NET/OneOrMany{T}.cs
+++ b/Source/Schema.NET/OneOrMany{T}.cs
@@ -4,6 +4,7 @@ namespace Schema.NET
     using System.Collections;
     using System.Collections.Generic;
     using System.Linq;
+    using System.Runtime.CompilerServices;
 
     /// <summary>
     /// A single or list of values.
@@ -15,8 +16,7 @@ namespace Schema.NET
         : IReadOnlyCollection<T>, IEnumerable<T>, IValues, IEquatable<OneOrMany<T>>
 #pragma warning restore CA1710 // Identifiers should have correct suffix
     {
-        private readonly List<T> collection;
-        private readonly T item;
+        private readonly T[] collection;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="OneOrMany{T}"/> struct.
@@ -24,8 +24,16 @@ namespace Schema.NET
         /// <param name="item">The single item value.</param>
         public OneOrMany(T item)
         {
-            this.collection = null;
-            this.item = item;
+            if (item is null || (item is string itemAsString && string.IsNullOrWhiteSpace(itemAsString)))
+            {
+                this.collection = null;
+                this.HasOne = false;
+            }
+            else
+            {
+                this.collection = new[] { item };
+                this.HasOne = true;
+            }
         }
 
         /// <summary>
@@ -33,8 +41,61 @@ namespace Schema.NET
         /// </summary>
         /// <param name="array">The array of values.</param>
         public OneOrMany(params T[] array)
-            : this(array is null ? null : new List<T>(array))
         {
+            if (array is null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+
+            if (array.Length > 0)
+            {
+                var items = new T[array.Length];
+                var index = 0;
+
+                if (typeof(T) == typeof(string))
+                {
+                    for (var i = 0; i < array.Length; i++)
+                    {
+                        var item = array[i];
+                        if (!string.IsNullOrWhiteSpace(item as string))
+                        {
+                            items[index] = item;
+                            index++;
+                        }
+                    }
+                }
+                else
+                {
+                    for (var i = 0; i < array.Length; i++)
+                    {
+                        var item = array[i];
+                        if (item != null)
+                        {
+                            items[index] = item;
+                            index++;
+                        }
+                    }
+                }
+
+                if (index > 0)
+                {
+                    if (index == array.Length)
+                    {
+                        this.collection = items;
+                    }
+                    else
+                    {
+                        this.collection = new T[index];
+                        Array.Copy(items, 0, this.collection, 0, index);
+                    }
+
+                    this.HasOne = index == 1;
+                    return;
+                }
+            }
+
+            this.collection = null;
+            this.HasOne = false;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -15,131 +16,130 @@ namespace Schema.NET
...
-        public OneOrMany(IEnumerable<T> collection)
-            : this(collection is null ? null : new List<T>(collection))
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="OneOrMany{T}"/> struct.
-        /// </summary>
-        /// <param name="list">The list of values.</param>
-        public OneOrMany(List<T> list)
```
RehanSaeed: Removing this is a breaking change. I'm ok with that but this will be a major version upgrade.

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -15,131 +16,130 @@ namespace Schema.NET
...
             }
         }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="OneOrMany{T}"/> struct.
         /// </summary>
-        /// <param name="list">The list of values.</param>
-        public OneOrMany(IEnumerable<object> list)
+        /// <param name="array">The array of values.</param>
+        public OneOrMany(params T[] array)
```
RehanSaeed: If we care about performance, perhaps we should add overloads for `Span<T>`?

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -15,131 +16,130 @@ namespace Schema.NET
...
+        /// <param name="collection">The list of values.</param>
+        public OneOrMany(IEnumerable<object> collection)
+            : this(collection?.Cast<T>())
+        {
         }
 
+        /// <summary>
+        /// Gets the number of elements contained in the <see cref="OneOrMany{T}"/>.
+        /// </summary>
+        public int Count => this.collection?.Length ?? 0;
```
RehanSaeed: I thought about doing this. Perhaps we should also implement `ICollection` if we've come this far?

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -15,131 +16,130 @@ namespace Schema.NET
...
         /// <value><c>true</c> if this instance has more than one value; otherwise, <c>false</c>.</value>
-        public bool HasMany => this.collection != null;
+        public bool HasMany => this.collection?.Length > 1;
 
         /// <summary>
         /// Performs an implicit conversion from <typeparamref name="T"/> to <see cref="OneOrMany{T}"/>.
         /// </summary>
         /// <param name="item">The single item value.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
```
RehanSaeed: What do you think of implementing the methods to do away with these warnings?

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -291,26 +287,6 @@ public bool Equals(OneOrMany<T> other)
...
-        }
-
-        /// <summary>
-        /// Checks whether the generic T item is a string that is either null or contains whitespace.
-        /// </summary>
-        /// <returns>
-        /// Returns true if the supplied item is a string that is null or contains whitespace.
-        /// </returns>
-        private static bool IsStringNullOrWhiteSpace(T item) => item.GetType() == typeof(string) && string.IsNullOrWhiteSpace(item as string);
+        public override int GetHashCode() => this.collection is null ? 0 : HashCode.OfEach(this.collection);
```
RehanSaeed: This can be simplified to `HashCode.OfEach(this.collection)`. `OfEach` handles null collections.

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: Yeah - its binary breaking though `List<T>` would still work to the `IEnumerable<T>` constructor if an application is recompiled.

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: Yeah - I was even thinking about using `Span<T>` internally instead of even the array. Unfortunately that would require the `OneOrMany` struct to be a `ref`.

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: Yeah, that's a good idea.

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: Yeah, implementing those methods shouldn't be an issue. [Looking up the docs](https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2225?view=vs-2019), it sounds like we just need `FromX` methods though it is vague whether those are meant to be static or what we would call it in the case of `T` - `FromT`?

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: Going back over this - we already implement `IReadOnlyCollection<T>` and as `OneOrMany` is immutable, many over the implemented interface methods on `ICollection<T>` would be no-ops or need to throw exceptions.

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: Going over this one again - I'm not sure we can implement good names that satisfy the rules.

For `T[]`, it wants the method named either `To` or `FromOneOrMany`. The former is just bad while the latter seems to imply a method on `T[]` itself - I don't think extension methods count to this analyzer rule.

For `T`, it wants the method named either `ToT` or `FromOneOrMany`. Now the former is slightly better than it was above but the latter will now be in conflict with above as two methods can't only be different by their return type.

When casting from a `List<T>` to `OneOrMany<T>`, we hit another problem. To have a method actually be `FromList`, it would make sense to be static on `OneOrMany<T>`. There is a conflicting rule that says static methods should not be on generic types. We could have `FromList` as a non-static method but that means we need a `OneOrMany<T>` to begin with.

I'm not sure what we want to do to satisfy the analyzer's specific naming requirements. In the mean time though, I did add decently named methods `ToArray` and `FirstOrDefault` which is what two of the implicit conversion use. While named like their LINQ counterparts, they are tailored for our data structure.

Source/Schema.NET/OneOrMany{T}.cs
RehanSaeed: Yes I agree about `ICollection<T>`. I think we could do `ICollection` for `CopyTo` but not a massive improvement.

Source/Schema.NET/OneOrMany{T}.cs
RehanSaeed: That must be why I didn't do it in the first place then. I guess we'll leave it as is then?

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -238,6 +238,38 @@ public IEnumerator<T> GetEnumerator()
...
+                return Array.Empty<T>();
+#endif
+            }
+        }
+
+        /// <summary>
+        /// Returns the first item from <see cref="OneOrMany{T}"/> or default <typeparamref name="T"/>.
+        /// </summary>
+        /// <returns>The first element in the <see cref="OneOrMany{T}"/></returns>
+        public T FirstOrDefault() => this.Count > 0 ? this.collection[0] : default;
```
RehanSaeed: We'll get this from `System.Linq` because we implement `IEnumerable<T>`. Do you still think it's worth implementing?

Turnerj: Yeah - I really only implemented it to try and satisfy the linting rule about having methods for implicit actions. It would be faster than [the default `FirstOrDefault`](https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/First.cs#L43-L77) implementation but at the same time, probably not actually worth the effort of keeping if we don't aim for the other linting rules. I'll remove it in another commit soon.

Source/Schema.NET/OneOrMany{T}.cs
Turnerj: `CopyTo` might be interesting - will look into it.

Source/Schema.NET/OneOrMany{T}.cs
RehanSaeed: Cool. I'm happy for this to be squash merged after this change. Feel free to do so.

PR #119, opened 2019-12-24
Faster instance creation for ValuesJsonConverter

Turnerj:
It turns out `Activator.CreateInstance` is slow relative to the rest of the serialization process.

image
In that image, you can see it taking a little over 30% of the deserialization time.

Using a technique from [JSON .NET](https://github.com/JamesNK/Newtonsoft.Json/blob/cdf10151d507d497a3f9a71d36d544b199f73435/Src/Newtonsoft.Json/Utilities/ExpressionReflectionDelegateFactory.cs#L44-L58) (with a similar technique in [System.Text.Json](https://github.com/dotnet/corefx/blob/e99ec129cfd594d53f4390bf97d1d736cff6f860/src/System.Text.Json/src/System/Text/Json/Serialization/ReflectionEmitMemberAccessor.cs#L15-L56)), the idea is to create a custom constructor delegate to avoid the overhead of finding and calling the constructor through reflection.

image
In this image, we can see the large drop in both the `System.Private.CoreLib` module as well as the line percentage dropping down to ~7% of the deserialization time.

With benchmarks, we can see the difference in performance and allocations as well. Specifically with the Book benchmark:

*Before*

|      Method |       Runtime |     Mean |    Error |   StdDev |   Median |      Min |      Max |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |-------------- |---------:|---------:|---------:|---------:|---------:|---------:|--------:|------:|------:|----------:|
|   Serialize |    .NET 4.7.2 | 205.1 us | 13.57 us | 28.33 us | 190.9 us | 183.1 us | 267.9 us | 15.3809 |     - |     - |  47.49 KB |
| Deserialize |    .NET 4.7.2 | 361.1 us |  6.54 us |  5.47 us | 360.4 us | 350.1 us | 371.7 us | 34.6680 |     - |     - |  107.3 KB |
|   Serialize | .NET Core 3.1 | 177.1 us |  3.10 us |  2.59 us | 177.2 us | 173.7 us | 181.0 us | 15.3809 |     - |     - |  47.27 KB |
| Deserialize | .NET Core 3.1 | 290.0 us |  4.08 us |  3.82 us | 290.0 us | 281.4 us | 295.7 us | 33.6914 |     - |     - |    104 KB |

*After*

|      Method |       Runtime |     Mean |    Error |   StdDev |   Median |      Min |      Max |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------ |-------------- |---------:|---------:|---------:|---------:|---------:|---------:|--------:|------:|------:|----------:|
|   Serialize |    .NET 4.7.2 | 206.4 us | 13.75 us | 27.47 us | 191.3 us | 182.7 us | 261.2 us | 15.3809 |     - |     - |  47.49 KB |
| Deserialize |    .NET 4.7.2 | 176.3 us |  1.98 us |  1.85 us | 176.9 us | 172.7 us | 178.4 us | 26.6113 |     - |     - |  81.84 KB |
|   Serialize | .NET Core 3.1 | 181.5 us |  3.14 us |  2.93 us | 181.5 us | 174.0 us | 185.6 us | 15.3809 |     - |     - |  47.27 KB |
| Deserialize | .NET Core 3.1 | 155.5 us |  1.48 us |  1.38 us | 155.4 us | 152.7 us | 158.1 us | 25.6348 |     - |     - |   79.1 KB |

As you can see, deserialization times are cut in half while also allocating ~24% less.

The constructor delegate cache is built per each object type (eg. `OneOrMany<string>` or `Values<string, Uri>`) with every type getting an entry. For a one-off serialization of a type, this would actually use more and likely take longer as each constructor gets generated for every type on every property being serialized. The benefits to this change come more with the frequency of deserializations.

-----

With the update to the benchmark runtime moniker, I found there to be a strange performance measuring issue when there is a mix of .NET 3.1 target in the project file with a runtime moniker targeting 3.0 while not having 3.0 installed (.NET Core was performing far worse than .NET Framework). I figure let's just make the project target and the benchmark runtime moniker be the same.

### DIFF
--- a/Source/Schema.NET/Schema.NET.csproj
+++ b/Source/Schema.NET/Schema.NET.csproj
@@ -45,6 +45,10 @@
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="16.4.43" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.1' Or '$(TargetFramework)' == 'net461'">
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+  </ItemGroup>
 
   <ItemGroup Label="Analyzer Package References">
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" PrivateAssets="all" Version="2.9.8" />
--- a/Source/Schema.NET/ValuesJsonConverter.cs
+++ b/Source/Schema.NET/ValuesJsonConverter.cs
@@ -19,8 +19,6 @@ namespace Schema.NET
     {
         private static readonly TypeInfo ThingInterfaceTypeInfo = typeof(IThing).GetTypeInfo();
         private static readonly Dictionary<string, Type> BuiltInThingTypeLookup = new Dictionary<string, Type>(StringComparer.Ordinal);
-        private static readonly ConcurrentDictionary<Type, Func<IEnumerable<object>, object>> TypeInstantiationMap = new ConcurrentDictionary<Type, Func<IEnumerable<object>, object>>();
-        private static readonly ParameterExpression[] ConstructorParameterExpression = new[] { Expression.Parameter(typeof(IEnumerable<object>)) };
 
         static ValuesJsonConverter()
         {
@@ -73,7 +71,7 @@ namespace Schema.NET
                 throw new ArgumentNullException(nameof(serializer));
             }
 
-            var dynamicConstructor = GetDynamicConstructor(objectType);
+            var dynamicConstructor = FastActivator.GetDynamicConstructor<IEnumerable<object>>(objectType);
 
             if (reader.TokenType == JsonToken.StartArray)
             {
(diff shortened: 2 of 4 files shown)


Turnerj:
As far as deserialization goes, there are probably two more low-hanging fruit in the library:

1. Optimising the `IEnumerable<object>` path in `Values` to not allocate 4 lists if they aren't all necessary (I've got a few ideas I need to test out with that)

2. Optimising the single-item path for `Values` and `OneOrMany` when coming from `ValuesJsonConverter`. Currently the converter simplifies itself and allocates an array for one item. In `OneOrMany` as an example, it allocates a second array for filtering - that is an OK compromise for a many items but results in one-more-than-necessary allocations for a single item. Really, both types could do with a type `object` constructor - need to see if it is worth it though.

Source/Schema.NET/ValuesJsonConverter.cs
```diff
@@ -169,6 +173,31 @@ public virtual void WriteObject(JsonWriter writer, object value, JsonSerializer
             serializer.Serialize(writer, value);
         }
 
+        private static Func<IEnumerable<object>, object> GetDynamicConstructor(Type objectType)
```
RehanSaeed: Perhaps this can be pulled out into another class. Maybe something like `FastActivator.CreateInstance`.

RehanSaeed:
I discovered the same thing some time ago. I also discovered that the generic `new()` constraint uses `Activator.CreateInstance` under the hood, so generic code can also be speeded up. I ended up using this implementation that the great Jon Skeet actually pointed me at:

https://github.com/Dotnet-Boxed/Framework/blob/master/Source/Boxed.Mapping/Factory.cs

Very similar to yours. I raised an issue in corefx at the time for `new()` generic constraing to be speeded up by avoiding the use of `Activator.CreateInstance` but it hasn't gained much traction.

Turnerj:
I've moved the object activation code out to a separate class like you suggested and extended it a little to support custom parameter types (which will be useful for a future PR around faster single-item deserialization).

Source/Schema.NET/Schema.NET.csproj
```diff
@@ -45,6 +45,10 @@
     <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="16.4.43" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
   </ItemGroup>
+  
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.1' Or '$(TargetFramework)' == 'net461'">
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
```
Turnerj: This dependency is required for the constructor delegate lookup in `FastActivator` though is only needed in two targets.

RehanSaeed:
Thanks for that.

PR #120, opened 2019-12-25
On-demand List allocation in Values<> type

Turnerj:
Previously we would allocate (up to) 4 lists in the `IEnumerable<object>` constructor on `Values<>`. This is fine when we have all types supported by `Values<>` present in the JSON we are reading but is overkill when we have less than that.

This change is effectively just-in-time list construction where we only instantiate that type when we actually have an item of that type.

According to my benchmarks, there is a 1.5% decrease in allocations - likely as a best case. As a worst case (a `Values4` has all 4 types present in JSON), the allocations should be the same.

### DIFF
--- a/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
+++ b/Source/Schema.NET/Values{T1,T2,T3,T4}.cs
@@ -101,40 +101,60 @@ namespace Schema.NET
                 throw new ArgumentNullException(nameof(items));
             }
 
-            var items1 = new List<T1>();
-            var items2 = new List<T2>();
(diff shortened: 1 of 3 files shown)


Turnerj:
There is an interesting quirk with this change - it now shows the problem I mentioned in #116 more obviously. With that in mind, this now depends on #116 - once that is merged, then I can rebase this on that and we can try again!

RehanSaeed:
Happy to take this. Just got some failing tests.

Turnerj:
I've rebased this on top of master now that the refactoring of `OneOrMany` is now merged - now the tests are passing.

Issue #123, opened 2020-01-15
Microsoft.VisualStudio.Threading.Analyzers gets added to dependent projects


csdecrypt:
### Describe the feature
Hi

The analyzer Microsoft.VisualStudio.Threading.Analyzers is automatically added to projects that use Schema.NET.

Altough i really like that analyzer, i'd rather not force consumers of this library to use it too.

(Src: https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#dependent-projects)

Turnerj:
Fixed in #124

RehanSaeed:
Thank you. I guess I missed that.

RehanSaeed:
I'll do a release soon.

RehanSaeed:
Released in https://github.com/RehanSaeed/Schema.NET/releases/tag/Schema.NET.7.0.1. CI builds will push to NuGet soon.

PR #124, opened 2020-01-15
set privateassets for Microsoft.VisualStudio.Threading.Analyzers

csdecrypt:
Addresses issue https://github.com/RehanSaeed/Schema.NET/issues/123

Setting PrivateAssets="All" should do the trick 
(Src: https://docs.microsoft.com/en-us/visualstudio/code-quality/use-roslyn-analyzers?view=vs-2019#dependent-projects)


Turnerj:
Looks good to me - thanks for the PR!

PR #128, opened 2020-01-30
Use global.json for sdk version

kirkone:
- Add a global.json
- Make use of the new feature that 'Install .NET Core SDK' can use
  the sdk version from global.json

This will prevent developing against a different version than that one
used by the build pipeline.


RehanSaeed:
This is interesting. I've had a `global.json` in various projects at times. I've found that people complain they can't build the project because they don't have the exact version of the SDK I have in `global.json`. I hope the error message has improved.

I like this idea though. I wonder if the `global.json` file could be read using PowerShell Core and used in AppVeyor and GitHub Actions too. Also looks like the GitHub Actions task is going to add the same feature in future https://github.com/actions/setup-dotnet/issues/14.

Only comment I would suggest apart from the above is to add the file to the `Solution Items` folder in the solution explorer.

kirkone:
> Only comment I would suggest apart from the above is to add the file to the Solution Items folder in the solution explorer.

Of course, forgot this and will add this.

For AppVeyor and GitHub Actions maybe we can use `dotnet --version`. It will return the version string like it is written in the `global.json`. The Problem is it will error out when the required version is not installed.

With PowerShell something like that could be helpfull

```powershell
(Get-Content .\global.json -Raw | ConvertFrom-Json).sdk.version
```

Should I implement this in this PR? I did not geht in touch with AppVeyor and GitHub Actions but I can have a look.

RehanSaeed:
If you don't implement it, I will. The choice is yours.

Also, thank you for your contribution.

RehanSaeed:
Thanks for your contribution. Going to follow up with https://github.com/RehanSaeed/Schema.NET/pull/138 to do the same for AppVeyor. GitHub Actions has a PR to do the same, so waiting for that.

kirkone:
Oh, sorry did not see your last comment. Thanks for accepting this PR.

RehanSaeed:
Good spot, nice to have the .NET Core version in one file.

PR #143, opened 2020-03-18
Use MinVer NuGet package for Versioning NuGet packages

RehanSaeed:
- [MinVer](https://github.com/adamralph/minver) does versioning using git tags, git height, build number and git commit SHA all combined.
- Added `Directory.Build.props` to configure common `.csproj` properties.
- Added `Directory.Build.targets` to configure MinVer.


RehanSaeed:
I'm considering going all in with `Directory.Build.props` and adding another one under the `Source` directory to configure assembly signing, turning on xml doc generation and configuring source link. Perhaps even another one under `Tests` to add `xunit` etc.

Turnerj:
Looks good to me! Definitely cleans up a lot of double-ups having a global `Directory.Build.props`.

Regarding having additional ones under tests and source, I guess it depends whether there will be more projects under those folders or whether the idea is to bring all non-specific configuration out of the single `.csproj` files.

That said, I'd probably still do it because there is still a chance we might have additional projects later on - whether that be we have separate projects for the extended schema.org types and then require splitting the tests into separate projects for each. 🙂

RehanSaeed:
> Regarding having additional ones under tests and source, I guess it depends whether there will be more projects under those folders or whether the idea is to bring all non-specific configuration out of the single .csproj files.

I guess it doesn't matter in this one. I have other projects where it will. I try to keep them all consistent though.

RehanSaeed:
Turns out you have to [add a nasty](https://github.com/Microsoft/msbuild/issues/2456) `<Import>` if you want to use multiple nested `Directory.Build.props` files. I'll do it in another PR.

PR #146, opened 2020-03-20
Add Multiple nested Directory.Build.props

RehanSaeed:


Directory.Build.props
```diff
@@ -21,6 +21,7 @@
 
   <ItemGroup Label="Package References">
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" PrivateAssets="All" Version="2.9.8" />
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" PrivateAssets="All" Version="1.0.0" />
```
RehanSaeed: Should have done this in the last PR.

RehanSaeed:
Not sure how I feel about this one. The tests is less useful I think. You could conceivably want to use a different test framework for example.

RehanSaeed:
Tests fail when you use another `Directory.Build.props`, so removed that one.

Directory.Build.props
Turnerj: Ahhhh - yeah, missed that when I went through the files. Good idea!

Turnerj:
I haven't had a case myself where I've wanted to mix and match test frameworks across multiple test projects in a single solution.

It does kinda suck that `Directory.Build.props` doesn't automatically import/merge from parent directories - I really thought that's how it worked.

Your changes look good to me!

PR #175, opened 2020-06-04
Tweak release drafter template and categories

RehanSaeed:
@dependabot-bot marks it's PR's with the `dependencies` label. I marked these as 'New Features' in release drafter. However, unit test NuGet package updates should show up in 'Maintenance' instead. This PR will make it so that each and ever PR will require manual labelling, so they appear in the correct section.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -6,7 +6,6 @@ categories:
   - title: '🚀 New Features'
     labels:
       - 'enhancement'
-      - 'dependencies'
   - title: '🐛 Bug Fixes'
     labels:
       - 'bug'
@@ -25,10 +24,8 @@ version-resolver:
       - 'patch'
   default: patch
 template: |
-  ## Changes
-
   $CHANGES
 
-  ## Contributors
+  ## 👨🏼‍💻 Contributors
 
   $CONTRIBUTORS


Turnerj:
Looks like the new dependabot integration in GitHub might allow us to have specific labels set for dependencies depending on the folder path.

See here about being able to configure labels: https://help.github.com/en/github/administering-a-repository/customizing-dependency-updates#about-customizing-dependency-updates
And here about having settings configured to a specific directory: https://help.github.com/en/github/administering-a-repository/enabling-and-disabling-version-updates#example-dependabotyml-file

Their example doesn't specifically say you can have the one package manager listed twice however with how they configured the YML, it looks like it might support it. This would mean then we could have something like this:

```yml
version: 2
updates:
  - package-ecosystem: "nuget"
    directory: "/Source/"
    labels:
      - dependencies
      - minor
    schedule:
      interval: "daily"

  - package-ecosystem: "nuget"
    directory: "/Tests/"
    labels:
      - dependencies
    schedule:
      interval: "weekly"
```

This is still just a theory at the moment however it looks like it might work.

RehanSaeed:
Oh cool, I didn't know this was possible. A few questions:

1. I'd really love this to be more automatic and handle the `maintenance` label for me but I don't see an easy way to do that.
2. If a dependency has it's patch number updated and we take that update, should we increment the `patch` or `minor` version? Just wondering about the `minor` label. Do you intend that to be a better default?

This is not going to be foolproof but I guess it might give a better defaults and save some time but also wondering if it's worth it for the `minor` label.

Turnerj:
Yeah, while we would have some extra control like being able to divide "test" package labels vs "source" package labels, we don't get the fine details of whether the label is for patch updates, minor version updates or major version updates.

For my example, `minor` was just the first thing that came to my head. Its a hard one where I could see `patch` being used for our updates where any `minor` or `patch` versions increment for our dependencies. If `minor` is meant to be "new features & fixes but backwards compatible", we may only be using it on our side for fixes so going up a `minor` version seems unnecessary.

Really our versions numbers ideally would go up on our changes alone. If we add new features in a backwards compatible way, that's `minor`. If we patch a problem, that's `patch`. If we have a dependency that goes up a major version and we are dependent on it, that might be the only time we are "forced" to jump a major version ourselves.

RehanSaeed:
In that case, I'm not sure adding this would be too valuable. We're always going to have to give dependabot a bit of thought and mess with it's labels. Nice to know this exists though.

PR #199, opened 2020-08-28
Workaround bug when pushing to GitHub Packages

RehanSaeed:
See https://github.com/NuGet/Home/issues/9775


Turnerj (approved):
Just read through that linked issue - that sucks but yeah, hopefully this workaround makes things a little less annoying.

Issue #197, opened 2020-08-21
URL for jsonld from schema.org changed?


MEmanuelsson:
I just downloded the source code for Schema.NET with the purpose of trying to see if I can find a way to solve this issue https://github.com/RehanSaeed/Schema.NET/issues/42, if possible. However running the project Schema.NET.Tool results in the application crashing since calling https://schema.org/version/latest/all-layers.jsonld results in status code 404 Not Found. 

Looking at https://schema.org/docs/developers.html it seems like this url has changed, probably to https://schema.org/version/latest/schemaorg-current-https.jsonld?

I tried changing to this url and that way I get information from schema.org but that then results in the following exception:

System.Exception: Check that a new primitive type has not been added.
   at Schema.NET.Tool.Services.SchemaService.SetParentAndChildren(List`1 classes) in C:\Source\Schema.NET\Tools\Schema.NET.Tool\Services\SchemaService.cs:line 180
   at Schema.NET.Tool.Services.SchemaService.GetObjectsAsync() in C:\Source\Schema.NET\Tools\Schema.NET.Tool\Services\SchemaService.cs:line 53
   at Schema.NET.Tool.Program.ExecuteAsync() in C:\Source\Schema.NET\Tools\Schema.NET.Tool\Program.cs:line 65
   at Schema.NET.Tool.Program.Main() in C:\Source\Schema.NET\Tools\Schema.NET.Tool\Program.cs:line 45

Any input here would be appreciated.
Best regards
Martin

MEmanuelsson:
I would guess that this pull request is where the change was made in the schema.org repo, https://github.com/schemaorg/schemaorg/pull/2654. Also announced on twitter as it seems: https://twitter.com/danbri/status/1283767294267731971

RehanSaeed:
Thanks @MEmanuelsson for the detective work!

Fixed in https://github.com/RehanSaeed/Schema.NET/pull/202.

RehanSaeed:
Fix has gone in. Will push a new release soon.

PR #202, opened 2020-09-03
Update Schema & Use HTTPS URL's

RehanSaeed:
- Updated the schema from schema.org.
- Fixed https://github.com/RehanSaeed/Schema.NET/issues/197.
- Schema.org provide a choice of new URL's. I've picked the one that prefixes all schema's with `https` instead of `http` and also updated all tests to match. This is a breaking change.
- Fixed a bug where a type with a single parent that was a pending type would be omitted even though it might have other parent types that were not pending.
- Pushing NuGet package to GitHub Actions now works.
- Updated the .NET Core SDK to 3.1.403.
- Updated `.editorconfig` file.


Turnerj (approved):
Looks good to me!

PR #209, opened 2020-09-28
Use latestMajor rollforward .NET SDK Policy

RehanSaeed:


RehanSaeed:
dotnet-install.sh Fails When global.json Contains rollForward=latestMajor
https://github.com/dotnet/sdk/issues/10650

PR #218, opened 2020-11-02
Use latestMajor .NET SDK

RehanSaeed:


RehanSaeed:
This will work on AppVeyor when the latest dotnet-install.sh file is released.

PR #226, opened 2020-11-17
Upgrade to .NET Core SDK 5.0.100

RehanSaeed:
- Upgrade to .NET Core SDK 5.0.100.
- Dropped support for `netstandard1.1`.
- Update the `.editorconfig` file which has changed in a breaking way (https://github.com/RehanSaeed/EditorConfig/pull/42) to support VS 16.8.
- Enable `EnforceCodeStyleInBuild` (This is awesome).
- Fix all the new warnings.
- Switch from `is object` to `is not null`. This has been on my mind for a long time. There is a lot more that could be done but getting some C# 9 in early.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,5 @@
-# Version: 1.6.2 (Using https://semver.org/)
-# Updated: 2020-11-02
+# Version: 2.0.0 (Using https://semver.org/)
+# Updated: 2020-11-15
 # See https://github.com/RehanSaeed/EditorConfig/releases for release notes.
 # See https://github.com/RehanSaeed/EditorConfig for updates to this file.
 # See http://EditorConfig.org for more information about .editorconfig files.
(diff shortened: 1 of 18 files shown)


Turnerj (approved):

Turnerj:
Looks good to me! Wasn't sure that C# 9 features were going to be supported by older runtimes (eg. Framework) as I remember seeing something saying that new C# features were going to coincide with new .NET releases but it is pretty cool to find out that at least things like `is not null` will be.

While this is not directly relevant to this issue, thought I would share something that I found out recently with some Microsoft.* 5.0.0 dependencies: https://github.com/Azure/azure-functions-vs-build-sdk/issues/472
This has nothing to do with us targeting .NET 5 in our tool/tests or using the SDK but if we were to update the wrong assemblies, it may cause issues under particular circumstances.

PR #246, opened 2021-02-15
Add NuGet package cache

RehanSaeed:
GitHub Actions - https://github.com/actions/cache/blob/main/examples.md#c---nuget
Azure Pipelines - https://docs.microsoft.com/en-us/azure/devops/pipelines/release/caching?view=azure-devops#netnuget
Appveyor - https://www.appveyor.com/docs/build-cache/ and https://help.appveyor.com/discussions/questions/31063-separate-cache-statements-per-operating-system

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,6 +37,12 @@ jobs:
       shell: pwsh
     - name: 'Install .NET Core SDK'
       uses: actions/setup-dotnet@v1
+    - uses: actions/cache@v2
+      with:
+        path: ~/.nuget/packages
+        key: ${{ runner.os }}-nuget-${{ hashFiles('**/*.csproj') }}-${{ hashFiles('**/*.props') }}
+        restore-keys: |
+          ${{ runner.os }}-nuget-
     - name: 'Dotnet Tool Restore'
       run: dotnet tool restore
       shell: pwsh
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -17,6 +17,9 @@ environment:
   # Set the build number in MinVer.
   MINVERBUILDMETADATA: build.%APPVEYOR_BUILD_NUMBER%
 
+cache:
+- packages -> ~/.nuget/packages
+
 build_script:
 # https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script
 - pwsh: |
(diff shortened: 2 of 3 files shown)


RehanSaeed:
Blocked by bugs in NuGet: https://github.com/NuGet/Home/issues/8465

Turnerj:
Hey, FYI, you might not need the packages lock file. I've had some success with NuGet caching on Azure DevOps with this:

```yml
  - task: Cache@2
    displayName: Cache NuGet Packages
    inputs:
      key: 'nuget | "$(Agent.OS)" | **/*.csproj'
      restoreKeys: |
         nuget | "$(Agent.OS)"
      path: $(NUGET_PACKAGES)
```

It does mean _any_ changes to a CS Project file throw out the cache though I don't think that's too bad. Not sure how this approach would translate to GitHub Actions (I assume pretty well but who knows) and AppVeyor (no idea).

Turnerj:
Oh, I guess it doesn't handle `Directory.Build.props` but even then, could probably toss that into the file expression with `**/Directory.Build.props`.

RehanSaeed:
Thats interesting.

Given we always hard code NuGet package version numbers unlike in NPM where we often use latest patch versions `"^1.0.0"`, I've been wandering what the advantage of the lock files is in the first place?

Turnerj:
I _think_ perhaps the lock file is more useful for full apps where you might have multiple dependencies with their own dependencies. Their dependencies may have different versions of the same library requested. Even with this said though, I'm pretty sure that is when the compiler yells at you for having mixed versions and requests you install the appropriate version directly.

RehanSaeed:
It seems that adding caching on AppVeyor actually slows the build down and I've seen similar comments about Azure Pipelines.

Also, I'm not sure what to set the cache `path` to on Ubuntu and MacOS. I'm using `~/.nuget/packages`. The strange thing is this seems to succeed on GitHub Actions and fail on Azure Pipelines.

Starting to wonder if this is more trouble than it's worth.

Turnerj:
Yeah, I've had mixed performance with caching. On one hand, restores are legit faster. On the other, the time it takes to upload any package changes ends up taking about as long as not having it. In my application, the packages total like 600MB which seems pretty crazy.

For my case, I am building on a Ubuntu VM via Azure DevOps and I use this for the path to NuGet: `$(Agent.TempDirectory)/.nuget/packages`

This is set as the `NUGET_PACKAGES` environment variable which I believe is actually used by `dotnet restore` etc automatically.

I think the thing that makes it a bit more annoying in this case is now having 3 different CI's. They are definitely useful if something is weird or unstable with one or another but when it comes time to change configuration behaviour, you need to not only replicate it 3 times but factor in discrepancies between them.

RehanSaeed:
**The Promise** - Cache NuGet packages on your CI build for faster 'dotnet restore' times by enabling `RestorePackagesWithLockFile`.
**The Reality**: Year old bugs make lock files in dotnet unusable and caching files in CI builds make it slower.

Abandoned....

Issue #127, opened 2020-01-29
Add C# 8 Nullable Reference Types


RehanSaeed:
### Describe the feature

Add [C# 8 Nullable Reference Types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references).

PR #252, opened 2021-02-27
Convert Schema classes generation to use source generators

### LINKED ISSUE #227: Convert Schema.NET.Tool From Newtonsoft.Json to System.Text.Json
The `SchemaRepository` uses a `JsonConverter` called `SchemaPropertyJsonConverter` using Newtonsoft.Json. It uses it to deserialize a response from a HTTP client.

Turnerj:
Switches from the previous file-based system for Schema classes to use source generators. Fixes up a bunch of tooling bugs along the way thanks to unknown/unexpected changes in the JSON files we consume.

Also closes #227 because why not! This was a fun issue because the source generator struggled to find System.Text.Json until I did some weird referencing for it - see `Schema.NET.Tool.csproj` for what I mean. I think though this would have happened for Newtonsoft.Json too.

At the point of time of writing this description, it does generate the classes but am having some issues referencing them. Some of our existing code is commented out, some of the old generation code still remains etc.

There is a relatively large lag time when building, about 13 seconds on my machine, which most of it comes from processing the JSON files. The actual generation of the final source is only 100ms. It only takes a second or two to download the JSON so it seems to come down to how some of the nested property loops work. We will want to speed this up to make the process easier to use.

Debugging the source generator is a pain as well. I currently have settings in the `Schema.NET.csproj` which have it create the raw files in the `obj` directory which is helpful. To actually debug the source generator, you need a snippet like the following in it:

```csharp
#if DEBUG
if (!System.Diagnostics.Debugger.IsAttached)
{
    System.Diagnostics.Debugger.Launch();
}
#endif
```

You don't always want this there as it _seems_ to prompt you often, even when you're not actually building (might be some weird debugging issue I was having). Also that debugging it seems to crash some behind-the-scenes processes for Visual Studio.


Turnerj:
And it looks like I screwed up the commit I started from. I wrote `git pull upstream master` and not `git pull upstream main` - probably should update my local branches too! *doh*

Turnerj:
I've managed to improve processing performance by nearly 5x according to my profiling, primarily avoiding loops within loops within loops.
- Schema properties now use a hash map to find existing enumerations rather than iterate all schema classes (5ca53fd)
- Schema classes no longer search every single schema property to find theirs, it is performed once and looked up (09069a3)
- Schema classes now leverage a hash map to find valid parents rather than iterate all other schema classes (09069a3)

While there is probably more performance we can manage to extract, the main time now is just downloading the JSON files.

Turnerj:
I _think_ all the format warnings in the build relate to the interface generation, it has a blank line at the end before the closing brace.

Tools/Schema.NET.Tool/Services/SchemaService.cs
```diff
@@ -43,7 +43,7 @@ public async Task<IEnumerable<GeneratorSchemaObject>> GetObjectsAsync()
             var enumerations = new List<GeneratorSchemaEnumeration>();
             var classes = new List<GeneratorSchemaClass>();
             foreach (var schemaClass in schemaClasses
-                .Where(x => !x.IsPrimitive && !x.IsArchived && !x.IsMeta && !x.IsPending))
```
RehanSaeed, line 39: Are you sure we're not going to get pending types added in?

Tests/Schema.NET.Test/core/BookTest.cs
```diff
@@ -108,7 +108,7 @@ public class BookTest
                         "}," +
                         "\"expectsAcceptanceOf\":{" +
                             "\"@type\":\"Offer\"," +
-                            "\"availability\":\"https://schema.org/InStock\"," +
+                            "\"availability\":\"schema:InStock\"," +
```
RehanSaeed, line 111: Can you please explain the URL change here? Were we incorrectly handling enumeration types?

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
```diff
@@ -1,15 +1,27 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <OutputType>Exe</OutputType>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
RehanSaeed, line 4: Does it matter what target framework this is? Can't we just use the latest `net5.0`?

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
```diff
@@ -0,0 +1,198 @@
...
+
+    [Generator]
+    public class SchemaSourceGenerator : ISourceGenerator
+    {
+        private IEnumerable<GeneratorSchemaObject> SchemaObjects { get; set; }
+
+        public void Initialize(GeneratorInitializationContext context)
+        {
+#pragma warning disable IDE0022 // Use expression body for methods
+            Task.Run(async () =>
```
RehanSaeed, line 22: Why do we wrap this code in `Task.Run` and then synchronously wait for it?

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
```diff
@@ -0,0 +1,198 @@
...
+            {
+                interfaceImplements = $" : I{parentType.Name}";
+            }
+
+            var classModifiers = schemaClass.IsCombined ? " abstract" : string.Empty;
+            var classImplements = parentType is not null ? $" {parentType.Name}," : string.Empty;
+            var allProperties = schemaClass.Properties.OrderBy(x => x.Order).ToArray();
+
+            return
+$@"namespace Schema.NET
```
RehanSaeed, line 89: Will it be easy for us to spit out Pending schema classes to a Schema.NET.Pending project? I guess the generator would have to run twice, once for each project right?

RehanSaeed (commented):
Thanks @Turnerj, this looks like nice work.

I haven't messed with source generators much yet partly because the tooling is not ready. Is it bad enough that we should wait for better VS tooling?

There seem to be a few remaining warnings to fix. I suspect the generated code also has potential warnings, do these actually show up?

Tools/Schema.NET.Tool/Services/SchemaService.cs
Turnerj: I've moved that logic out to the SchemaRepository as it seemed to make sense having it control whether we wanted archived, meta and pending types. This is to also partially set up the path to supporting a separate "pending" schema library.

Tests/Schema.NET.Test/core/BookTest.cs
Turnerj: It seems that Schema.org changed enumeration values to be `schema:EnumerationName`. I updated our JSON converter to read the old HTTP/HTTPS versions as well as this new version. It writes the new version as that is what the JSON files from Schema.org provide. The test though does a exact string match so it needed to be updated to match the new value.

To be honest, I don't like that Schema.org changed it as I imagine it is a breaking change for anyone who stored the value. I think though we should use whatever value they are currently providing though.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
Turnerj: We have to use .NET Standard 2.0 because the generator runs in the frameworks for the target that uses the generator. Because we target .NET Framework and older .NET Core versions, we need use the most compatible target framework for the generator.

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
Turnerj: A combination of wanting to use the async/await pattern for the actual main logic while also trying to avoid potential deadlocks. Using `Task.Run` pushes it to the thread pool whereas async doesn't. I can give it another crack to remove/simplify it - this just happened to be what I had when I first got the source generator running at all.

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
Turnerj: I'm not 100% sure though I was trying to keep that in mind while building this. No doubt changes to the source generator will need to happen for support to a "pending schema" project. One of the things we will be able to do is put options in the target csproj file that the source generator can read. We can use that to pass options into the `SchemaService` to choose what classes to generate.

Turnerj:
> I haven't messed with source generators much yet partly because the tooling is not ready. Is it bad enough that we should wait for better VS tooling?

Honestly, for debugging the source generator, you need to be pretty patient and do some weird hacks for debugging. For working with the library that has a working source generator, it seems to be pretty easy going.

I think even with the debugging issues, the changes make understanding the generator logic easier while opening more possibilities. I think while we will have a rough patch for tooling to catch up, I think the source generator is a net positive.

> There seem to be a few remaining warnings to fix. I suspect the generated code also has potential warnings, do these actually show up?

The warnings don't show up in VS but you can see them via the command line using `dotnet build`. Should be able to fix them without too many issues.

Tests/Schema.NET.Test/core/BookTest.cs
RehanSaeed: Have you seen a doc or GitHub issue discussing this change? It looks really strange that they'd go and break everyone like this. If we want to be interoperable with others, consumers of this package will want to output the full URL's instead of the new format.

RehanSaeed:
> Honestly, for debugging the source generator, you need to be pretty patient and do some weird hacks for debugging. For working with the library that has a working source generator, it seems to be pretty easy going.
> 
> I think even with the debugging issues, the changes make understanding the generator logic easier while opening more possibilities. I think while we will have a rough patch for tooling to catch up, I think the source generator is a net positive.

Have you heard anything from the VS team about when they're going to add better support? It seems pretty hacky at the moment. I really like that we no longer checkin all of the types, which will make PR's going forward a lot simpler to navigate.

> The warnings don't show up in VS but you can see them via the command line using `dotnet build`. Should be able to fix them without too many issues.

That's not great, I've seen this occasionally but not sure what causes it. Ideally, we should get to the bottom of that.

Overall, this looks good. Tooling will get better. My main concern is the new enum format which we should get to the bottom of.

Tests/Schema.NET.Test/core/BookTest.cs
Turnerj: I think I worked out why the change has happened. In prior versions of the JSONLD file we download, they had the full URL. In the new one, they don't BUT they have something else added to the top most context:

image

The "schema" property on "@context" has the URL - I think the intention is to replace "schema:" with "https://schema.org/". Will fix that up and revert those "schema:" specific changes I've made.

Turnerj:
> Have you heard anything from the VS team about when they're going to add better support? It seems pretty hacky at the moment.

In VS 16.9 Preview 3, they added the ability to view generated files through solution explorer, that is one big step to helping debug source generators. https://docs.microsoft.com/en-us/visualstudio/releases/2019/release-notes-preview#NET-Productivity3

Generated source in solution explorer

Besides that, I did find that [a Roslyn dev suggests using a console application for debugging](https://twitter.com/davidwengier/status/1326267271749103616). We could do something similar, having a "Schema.NET.Generator.Debugger" console application that is similar to this: https://github.com/davidwengier/SourceGeneratorTemplate/blob/master/TestConsoleApp/Program.cs#L71

Don't know if we will have issues because of other dependencies like Newtonsoft.Json as this method seems to build an entire assembly from scratch rather than just add to an existing one.

RehanSaeed:
Since 16.9 has been released, I'm more comfortable with going with this.

Turnerj:
I just realised the tool previously included properties that are pending. For example, [`BankAccount` has 3 custom properties](https://github.com/RehanSaeed/Schema.NET/blob/main/Source/Schema.NET/core/BankAccount.cs), all of which [are pending on Schema.org](https://schema.org/BankAccount) but we have them generated.

Is this intentional? Should we be generating pending properties on our normal types? We even have a test against it here: https://github.com/RehanSaeed/Schema.NET/blob/30068b7b2205eccf3e794d5d99b2a0ad50e6a00c/Tests/Schema.NET.Test/MixedTypesTest.cs#L73-L83

Only noticed it while trying to "fix" my code that was meant to prevent pending types.

Turnerj:
Personally, I think it would be cleaner to not have pending properties in the main library. Basically we could have 2 libraries where one has all the non-pending classes/properties/enums and the other which has everything. I don't see another cleaner way of doing this.

RehanSaeed:
I like your idea to have all pending types in a separate library.

Yes, I'd like to automate the versioning. Perhaps the XML feed contains a version number we can run the GitHub actions from.

Turnerj:
So having pending _types_ in a separate library isn't a problem but how do we handle pending _properties_ or properties that can take a pending type? If the property is pending on a type that isn't pending, what do we do? If a property is updated to take a pending type, what do we do?

Splitting the assemblies covers us for pending types that extend existing types but doesn't cover us for pending properties or properties that are updated to take pending types. If we do 2 assemblies where one has no pending types/property and the other has everything (all pending and non-pending types and properties), the types won't be compatible across assemblies.

Does that make sense?

RehanSaeed:
It certainly does make sense. I think the least worst approach is to go with two assemblies containing the all relevant types. You'd have to choose from one assembly or the other. If we go with the shared assembly approach, we either end up with missing pending properties or pending properties on released types. Neither of which seems like a good idea to me.

Turnerj:
I've removed the now invalid test that was testing a pending property (that is now no longer generated). I also looked into the warnings I mentioned earlier, they don't seem to happen on Windows but they appear on the CI under Ubuntu and MacOS.

You can see it annotated on the changed files here: https://github.com/RehanSaeed/Schema.NET/pull/252/files#diff-d9f4bad47ff6976da03a740c615555119f094fa6550ea47ce813809aaffdff9f

The warnings are just "fix formatting" without any additional information. Visual Studio doesn't mention anything and looking at the annotated code, I can't see what the warning would be about.

Besides that, I'm happy with the changes at the moment. Getting this in sets us up well for a separate PR to work on adding a second assembly with all types (inc. pending).

RehanSaeed:
@Turnerj I'm seeing those warnings on all my repos. I'm not certain what they're all about. In theory they should fail the build since we treat warnings as errors but they don't. The warning code starts with `IDE`, so it's probably a warning caused by the `.editorconfig` file. So far, I've been putting it down to a .NET tooling bug since `.editorconfig` is still being worked on.

RehanSaeed:
If you run `dotnet build` from CI, you do get the warning and can get an idea of what it might be. Personally, I'm happy to fix the warning once .NET fixes it's tooling and the warning also shows up in VS.

Turnerj:
Shall we merge these changes in then?

RehanSaeed:
Sure! I guess we can release now, although I've been adding nullable support to all my repos and might do this one too to fix https://github.com/RehanSaeed/Schema.NET/issues/127.

PR #253, opened 2021-03-03
Fix VS 16.9 Warnings

RehanSaeed:
cc @Turnerj 

I think VS 16.9 switched on some new warnings from the `.editorconfig` file.


Turnerj:
Looks good to me!

RehanSaeed:
Also updated the .editorconfig file with new settings from https://github.com/RehanSaeed/EditorConfig/pull/49

PR #259, opened 2021-03-15
Enable C# 8 Nullable Reference Types

RehanSaeed:
Fixes https://github.com/RehanSaeed/Schema.NET/issues/127


RehanSaeed:
I spent some time starting this. I have fixed most issues (errors look worse than they are because the types can't be generated until the build is fixed) but there is some internal stuff that I'm unsure of. It seems that there are a lot of places where we are not handling potential nulls correctly.

@Turnerj Would you like to cast your eye over the last 10% of errors? Then I'll see if I can fix anything that's left.

Benchmarks/Schema.NET.Benchmarks/Core/BookBenchmark.cs
```diff
@@ -2,81 +2,79 @@ namespace Schema.NET.Benchmarks.Core
...
 
     public class BookBenchmark : SchemaBenchmarkBase
     {
-        [GlobalSetup]
-        public void Setup() => this.ConfigureBenchmark(new Book()
-        {
-            Id = new Uri("https://example.com/book/1"),
-            Name = "The Catcher in the Rye",
-            Author = new Person()
+        public override Thing InitialiseThing() =>
```
RehanSaeed, line 8: Made benchmarks more type safe which also fixes a bunch of null errors. Potentially, this could have been a separate PR.

Turnerj:
Yep, no problem!

Source/Schema.NET/ValuesJsonConverter.cs
```diff
@@ -307,18 +309,15 @@ private static bool TryProcessTokenAsType(JsonReader reader, Type targetType, ou
                         enumString = valueString;
                     }
 
-                    try
+                    if (Enum.TryParse(targetType, enumString, out result))
```
RehanSaeed, line 312: `Enum.TryParse` instead of try/catch. This could also be in another PR.

Source/Schema.NET/OneOrMany{T}.cs
```diff
@@ -278,11 +278,11 @@ public bool Equals(OneOrMany<T> other)
             }
             else if (this.HasOne && other.HasOne)
             {
-                return this.collection[0].Equals(other.collection[0]);
+                return Equals(this.collection![0], other.collection![0]);
```
Turnerj, line 281: Was the reason for changing to use `object.Equals(self, other)` to avoid checking directly if `this.collection[0]` is `null`?

Source/Schema.NET/ValuesJsonConverter.cs
Turnerj: Don't know why we didn't do this sooner!

Turnerj:
Nullable references support for the tool is done. Will look into the other bits left for the various `Values<>` types and then a few more in the `ValuesJsonConverter` later tonight.

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaEnumeration.cs
```diff
@@ -8,6 +8,11 @@ namespace Schema.NET.Tool.GeneratorModels
     public class GeneratorSchemaEnumeration : GeneratorSchemaObject
 #pragma warning restore CA1724 // Identifiers should conflict with namespaces
     {
+        public GeneratorSchemaEnumeration(string name, string layer)
```
RehanSaeed, line 11: I started to do this too and then reverted. In fact, I tried using `record` instead but got an error because the project is `netstandard2.0`.

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaObject.cs
```diff
@@ -2,8 +2,14 @@ namespace Schema.NET.Tool.GeneratorModels
...
     {
-        public string? Layer { get; set; }
+        public GeneratorSchemaObject(string name, string layer)
+        {
+            this.Name = name;
+            this.Layer = layer;
+        }
 
-        public string? Name { get; set; }
+        public string Layer { get; set; }
```
RehanSaeed, line 11: Should we remove the `set;` to make it immutable?

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaProperty.cs
```diff
@@ -7,16 +7,23 @@ namespace Schema.NET.Tool.GeneratorModels
...
+            this.Name = name;
+            this.JsonName = jsonName;
+        }
+
+        public GeneratorSchemaClass Class { get; }
 
         public string? Description { get; set; }
 
-        public string? JsonName { get; set; }
+        public string JsonName { get; set; }
```
RehanSaeed, line 24: Should we remove the `set;` for the values passed via the constructor?

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaObject.cs
```diff
@@ -2,8 +2,14 @@ namespace Schema.NET.Tool.GeneratorModels
 {
     public class GeneratorSchemaObject
     {
-        public string? Layer { get; set; }
+        public GeneratorSchemaObject(string name, string layer)
```
RehanSaeed, line 5: I think it would make more logical sense for the layer parameter to come first since it is higher up in the hierarchy.

Source/Schema.NET/DateTimeHelper.cs
```diff
@@ -18,8 +18,13 @@ internal static class DateTimeHelper
         /// </summary>
         /// <param name="input">The input string</param>
         /// <returns>True if the input string has an offset defined</returns>
-        public static bool ContainsTimeOffset(string input)
+        public static bool ContainsTimeOffset(string? input)
         {
+            if (input is null)
```
RehanSaeed, line 23: I made the methods in this helper handle nulls but I wonder if I should have handled null further up and then these can go back to be non-nullable.

Source/Schema.NET/OneOrMany{T}.cs
RehanSaeed: Yes. The alternative would be to check all the combinations `null==null`, `null==object`, `object==null` and `object==object`.

Source/Schema.NET/ValuesJsonConverter.cs
```diff
@@ -289,14 +291,14 @@ private static bool TryProcessTokenAsType(JsonReader reader, Type targetType, ou
                 }
                 else if (targetType.GetTypeInfo().IsEnum)
                 {
-                    string enumString;
-                    if (valueString.StartsWith(HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
+                    string? enumString;
+                    if (valueString is not null && valueString.StartsWith(HttpSchemaOrgUrl, StringComparison.OrdinalIgnoreCase))
```
RehanSaeed, line 295: As I mentioned elsewhere, perhaps a better way would be to check `valueString` for null in one place further up.

Tools/Schema.NET.Tool/GeneratorModels/ICloneable{T}.cs
```diff
@@ -1,7 +1,12 @@
-﻿namespace Schema.NET.Tool.GeneratorModels
+namespace Schema.NET.Tool.GeneratorModels
 {
     public interface ICloneable<T>
     {
         T Clone();
     }
+
+    public interface ICloneable<TOut, TIn>
```
RehanSaeed, line 8: I think you'll be getting a warning about putting two interfaces in one file.

Source/Schema.NET/ValuesJsonConverter.cs
RehanSaeed: Exactly! Performance win too!

Tools/Schema.NET.Tool/GeneratorModels/ICloneable{T}.cs
Turnerj: Part of me was thinking of dropping the interfaces entirely. We don't actually call through them directly as we have access to the concrete type.

Source/Schema.NET/DateTimeHelper.cs
Turnerj: This looks like it follows the behaviour of other `Try*` methods that we call in our `ValuesJsonConverter` so I think you made the right decision.

Source/Schema.NET/ValuesJsonConverter.cs
Turnerj: I think I found out why - that method doesn't exist in .NET Standard 2.0 😣

Turnerj:
Looks like all the nullability changes should be in now though am getting an error on .NET Standard 2.0 at the moment:

image

I think it is just Visual Studio being weird at the moment.

Source/Schema.NET/ValuesJsonConverter.cs
```diff
@@ -309,11 +309,20 @@ private static bool TryProcessTokenAsType(JsonReader reader, Type targetType, ou
...
 
+#if NETSTANDARD2_0 || NET472 || NET461
+                    try
+                    {
+                        result = Enum.Parse(targetType, enumString);
+                        success = true;
+                    }
+                    catch (Exception)
+#else
                     if (Enum.TryParse(targetType, enumString, out result))
```
Turnerj: This seemed like an OK compromise. Not a huge fan of using 3 targets like I am here though.

It did make me wonder, why do we target .NET 4.7.2 and .NET 4.6.1 directly when we also target .NET Standard 2.0?

Turnerj: I repeat this type of check here as well with my switched up way to generate a `ToString` for `PropertyValueSpecification`: https://github.com/RehanSaeed/Schema.NET/pull/259/commits/85dde5a5605bd2449358949a6306aaf537cf2d0b#diff-4170f976c7ab6112252917fe8fa2c6a2ce7bed515547e0e42e11ea4d67402f9dR47-R51

Turnerj:
What are your thoughts about disabling nullable reference types for our tests? Our tests are designed to check very specific conditions anyway and will fail on any exception. We might be just adding all the additional markup for it just for it to be mostly irrelevant.

RehanSaeed (commented):
The tests I've enabled nullable for have mostly involved liberal use of `!` and `?` with the occasional switch to immutable types. So nothing too onerous. We don't have many changes to tests currently but not sure if that'll change once everything is converted. Agree that it did feel mostly irrelevant, although I did manage to make the odd improvement.

Overall, I've been surprised at how little code has had to change to my other repos (this one seems the most complex) and nullable seems to have only made a marginal improvement.

Source/Schema.NET/PropertyValueSpecification.Partial.cs
```diff
@@ -17,57 +17,38 @@ public partial class PropertyValueSpecification
         /// </returns>
         public override string ToString()
         {
-            var stringBuilder = new StringBuilder();
+            var values = new List<string>();
```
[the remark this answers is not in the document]
Turnerj: I hadn't benchmarked it - when going through it to support nullable reference types, I liked the brevity of this solution. I would think in this case, the performance might be pretty close though _may_ allocate slightly more.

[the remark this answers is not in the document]
Turnerj: Yep, you're right, there is a performance loss with this method even though it is a few less `.First()` calls. Here is a benchmark I put together in LINQPad:

image

The "StringBuilderAfter" combines both methods - using `StringBuilder` with the way I have it skipping the extra `.First()` calls. Will implement this version in our code shortly.

Tests/Schema.NET.Test/MixedTypesTest.cs
```diff
@@ -44,7 +44,7 @@ public class MixedTypesTest
         [Fact]
         public void DeserializeObject_WithCollectionOfMixedTypes_ReturnsObjectWithBothTypes()
         {
-            var book = JsonConvert.DeserializeObject<Book>(this.json, TestDefaults.DefaultJsonSerializerSettings);
+            var book = JsonConvert.DeserializeObject<Book>(this.json, TestDefaults.DefaultJsonSerializerSettings)!;
 
             Assert.True(book.Author.HasValue);
```
RehanSaeed, line 49: Apparently `Assert.NotNull` supports nullable reference types, so if you assert that and use it's return value, you can avoid the exclamation here.

Tests/Schema.NET.Test/ThingTest.cs
```diff
@@ -147,15 +147,17 @@ public class ThingTest
                 AdditionalName = "B",
             });
 
-        private static void CompareEqual<T>(T a, T b)
+        private static void CompareEqual<T>(T a, T? b)
         {
-            Assert.Equal(a.GetHashCode(), b?.GetHashCode());
+            Assert.NotNull(a);
+            Assert.Equal(a!.GetHashCode(), b?.GetHashCode());
```
RehanSaeed, line 153: Same here about `Assert.NotNull`

RehanSaeed:
Build is green!

Tests/Schema.NET.Test/MixedTypesTest.cs
Turnerj: Had a look but it doesn't seem to be the case - `Assert.NotNull` returns voids.

image

Source/Schema.NET/ValuesJsonConverter.cs
Turnerj: Done! Created an `EnumHelper` static class with a `TryParse` method that copies the signature for `Enum.TryParse`.

RehanSaeed:
LGTM, are we ready to merge?

Turnerj:
Yeah, I'm happy with it!

PR #262, opened 2021-03-21
Source generator tweaks

Turnerj:
Various source generator tweaks and improvements:

- The comment field is now immutable (we were already refusing to parse any classes/enums/properties that didn't have one so it made sense to do)
- Optimizing the parsing of various properties depending on the type (eg. only parsing `schema:domainIncludes` for properties only)
- Deserializing the HTTP stream for the JSON rather than a string buffer (note the `HttpCompletionOption` which helps this) which can improve performance and allocations
- Removes the additional `Task.Run` in the generator's initializer
- Caches schema object data between different runs of the source generator for the different targets

Other plans:
- ~Investigate whether the source generator is triggered for every target (if so, see if we can cache the JSON)~
- Any other performance improvements in processing the JSON

Other notes:
- Found a weird quirk with #261 that it compiles with `dotnet build` but not `msbuild` or via Visual Studio. See dotnet/roslyn#52017 - the CI isn't affected as far as I can tell.

### DIFF
--- a/Tools/Schema.NET.Tool/Models/SchemaClass.cs
+++ b/Tools/Schema.NET.Tool/Models/SchemaClass.cs
@@ -9,8 +9,8 @@ namespace Schema.NET.Tool.Models
     {
         private static readonly Uri EnumerationId = new("https://schema.org/Enumeration");
 
-        public SchemaClass(Uri id, string label, string layer)
-            : base(id, label, layer)
+        public SchemaClass(string comment, Uri id, string label, string layer)
+            : base(comment, id, label, layer)
         {
         }
 
--- a/Tools/Schema.NET.Tool/Models/SchemaClass.cs
+++ b/Tools/Schema.NET.Tool/Models/SchemaClass.cs
@@ -9,8 +9,8 @@ namespace Schema.NET.Tool.Models
     {
         private static readonly Uri EnumerationId = new("https://schema.org/Enumeration");
 
-        public SchemaClass(string comment, Uri id, string label, string layer)
-            : base(comment, id, label, layer)
+        public SchemaClass(string layer, Uri id, string label, string comment)
+            : base(layer, id, label, comment)
         {
         }
 
--- a/Tools/Schema.NET.Tool/CustomOverrides/AddQueryInputPropertyToSearchAction.cs
+++ b/Tools/Schema.NET.Tool/CustomOverrides/AddQueryInputPropertyToSearchAction.cs
@@ -23,10 +23,7 @@ namespace Schema.NET.Tool.CustomOverrides
                 throw new ArgumentNullException(nameof(c));
             }
 
-            var property = new GeneratorSchemaProperty(c, "query-input", "QueryInput")
-            {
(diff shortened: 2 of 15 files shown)


Turnerj:
> Investigate whether the source generator is triggered for every target (if so, see if we can cache the JSON)

Confirmed - initialization and executing occur multiple times. Next check will be if they use the same instance of the generator and if they do, we can skip the initialization if the `SchemaObjects` property is not null.

**Example Log**
> Initialize: 21/03/2021 5:21:30 AM
Execute: 21/03/2021 5:21:30 AM
Initialize: 21/03/2021 5:21:39 AM
Execute: 21/03/2021 5:21:39 AM
Initialize: 21/03/2021 5:21:40 AM
Execute: 21/03/2021 5:21:40 AM
Initialize: 21/03/2021 5:21:40 AM
Execute: 21/03/2021 5:21:40 AM
Initialize: 21/03/2021 5:21:40 AM
Execute: 21/03/2021 5:21:40 AM

Turnerj:
> Next check will be if they use the same instance of the generator and if they do, we can skip the initialization if the `SchemaObjects` property is not null.

Unfortunately they use different instances _however_ as a static property, we can share data between them. The multi-target builds seem to overlap so adding some locking to it seems like a good idea.

Tools/Schema.NET.Tool/Models/SchemaClass.cs
```diff
@@ -9,8 +9,8 @@ public class SchemaClass : SchemaObject
     {
         private static readonly Uri EnumerationId = new("https://schema.org/Enumeration");
 
-        public SchemaClass(Uri id, string label, string layer)
-            : base(id, label, layer)
+        public SchemaClass(string comment, Uri id, string label, string layer)
```
RehanSaeed, line 12: Minor comment. It would be nice if the order of parameters in these constructos followed the hierarchy of data.

`string layer, Uri id, string label, string comment`

RehanSaeed:
Nice nice improvements. Do you think we can release a major version after this?

Tools/Schema.NET.Tool/Models/SchemaClass.cs
Turnerj: Happy to do that. While I know what hierarchy means, I'm not sure exactly how you're deriving it - like these properties exist in the same hierarchy of inheritance. Just want to understand how you want it.

I went alphabetical for this but felt weird to have comment first.

Turnerj:
We can, yeah. I do want to try and get pending types but that may be better after.

Tools/Schema.NET.Tool/Models/SchemaClass.cs
RehanSaeed: I think layer comes first because it groups many types together. The ID for a type makes sense next as it is the unique identifier. After that the label seems to make sense to me, although I can't explain why. After these three, I think going with alphabetical ordering is fine, unless it makes sense to group parameters together somehow.

RehanSaeed:
LGTM

PR #264, opened 2021-03-23
Further Nullable Improvements

RehanSaeed:
- Further nullable improvements.
- Also realized we were not using `Span<T>` when targeting .NET 5, so fixed that.
- Use `is not null`.


Turnerj (approved):
LGTM!

PR #266, opened 2021-03-29
Refining and restructuring the test cases

Turnerj:
**Changes**
- Type-specific tests (Book, MusicEvent, Website) are now in an "Examples" folder and namespace (the namespace change is useful when using the test explorer as they have their own expand/collapse node)
- Type-specific tests no longer have a dedicated `ToString` test (avoids issues with JSON property order) however a round trip conversion test (convert to JSON and back to object) still exists for each proving that the `ToString` method works.
- `MixedTypesTest` has moved to examples folder (mainly because the logic behind it is tested in `ValuesJsonConverterTest` however seemed worth keeping as an example)
- Core parts of `SerializationTest` have been moved into `ThingTest` with `SerializationTest` being deleted
- All tests use `SchemaSerializer` instead of directly using `JsonConvert` (this is the ideal way people will serialize our types and is what we directly call in `Thing.ToString()`)

Overall nothing overly major has changed one way or another but it should eliminate the issue in #100 where tests failed due to JSON order while making a few other bits a little clearer.


RehanSaeed:
Seems reasonable. If we're going to mess with tests, perhaps we can also choose a snapshot testing tool. There are a lot of options out there since snapshot testing in .NET seems to be in its infancy. I have started to evaluate some in https://twitter.com/RehanSaeedUK/status/1375470767475982336?s=20. So far, I'm leaning towards Verify.

Turnerj:
Hmmm, I was looking into Verify but I'm not sure exactly how it (or tools like it) would fit in with what we have. Like you mentioned in #100, what we have for a number of the tests are good examples for different things - that's fine, just not sure where tests using a snapshot testing tool help us.

Happy to give it a shot though might need to understand more where it fits in and how.

Tests/Schema.NET.Test/Examples/BookTest.cs
```diff
@@ -1,218 +1,213 @@
...
+            Assert.Equal(this.book.ToString(), SchemaSerializer.DeserializeObject<Book>(this.json)!.ToString());
+            Assert.Equal(SchemaSerializer.SerializeObject(this.book), SchemaSerializer.SerializeObject(SchemaSerializer.DeserializeObject<Book>(this.json)!));
+        }
+
+        [Fact]
+        public void Deserializing_BookJsonLd_ReturnsBook()
+        {
+            var book = SchemaSerializer.DeserializeObject<Book>(this.json)!;
+
+            Assert.Equal("The Catcher in the Rye", book.Name);
```
RehanSaeed, line 166: Here is an example of where we could use Verify. Wherever we have long lists of Asserts.

Tests/Schema.NET.Test/Examples/BreadcrumbListTest.cs
```diff
@@ -1,100 +1,95 @@
...
+            Assert.Equal(this.breadcrumbList.ToString(), SchemaSerializer.DeserializeObject<BreadcrumbList>(this.json)!.ToString());
+            Assert.Equal(SchemaSerializer.SerializeObject(this.breadcrumbList), SchemaSerializer.SerializeObject(SchemaSerializer.DeserializeObject<BreadcrumbList>(this.json)!));
+        }
+
+        [Fact]
+        public void Deserializing_BreadcrumbListJsonLd_ReturnsBreadcrumbList()
+        {
+            var breadcrumbList = SchemaSerializer.DeserializeObject<BreadcrumbList>(this.json)!;
+
+            List<IThing> things = breadcrumbList.ItemListElement;
```
RehanSaeed, line 77: Another example.

RehanSaeed:
Hmmm, I thought a snapshot tool would be useful to get rid of the JSON strings but now that I look at it, because we do serialize and deserialize tests, we need the JSON string. Ok, ignore me then.

Turnerj:
Any other thoughts with this PR? Happy to have it merged?

RehanSaeed:
LGTM

RehanSaeed (approved):

RehanSaeed:
Good time to release?

Turnerj:
Sounds good to me! Will do some more stuff with S.T.J soon and maybe open a PR to support pending types like we discussed.

PR #281, opened 2021-04-30
Add NuGet README.md

RehanSaeed:


RehanSaeed:
Correct yes, I should have linked that issue.

Turnerj:
FYI, it looks like .NET 6 RC2 will have the fix for the blocking issue: https://github.com/NuGet/Home/issues/10791#issuecomment-926218883

Issue #165, opened 2020-05-03
Add metadata for Schema.org Version


Turnerj:
### Describe the feature

I'm not 100% sure we actually get this from our data source but it would be nice having a program-accessible constant that refers to the Schema.org version that the code is currently using.

RehanSaeed:
Decent idea. It might be in one of the two JSON files we request.

Turnerj:
With our request to `https://schema.org/version/latest/all-layers.jsonld`, the `@id` property (currently) contains "https://schema.org/#8.0" so getting the version isn't too hard.

To get the version without a second HTTP request might be harder due to how the code in the tool handles the result - its designed to pass back only the schema classes.

Might need to think a little more about how I want to approach it. I know I've wanted to restructure the tool a bit to make working on it a little easier but I didn't want to overlap with potentially changing how the tool generates the classes (as like we've discussed before, source generators in .NET 5 might be the path forward).

PR #301, opened 2021-06-19
Improve source generator usage

Turnerj:
There are 3 main things this PR does:
- The source generator now uses the latest previews of System.Text.Json
_This helps work around other weird Visual Studio behaviour when compiling the code and versions etc_
- Adds the flags so we can debug source generators via Visual Studio
_This is an amazingly useful feature!_
- Reverts to using a file for schema generation
_Something wasn't sitting right with me that our builds were not deterministic - it was up to the state of the JSON file downloaded. I thought via a file might be faster too but so far haven't seen that be the case._

With that last point, I think we can still semi-automate the updating process for the JSON file I've embedded via scheduled GitHub Actions. This might also allow us to support #165 more easily.

In theory, none of these changes should affect the output of Schema.NET, just the build/debugging process.


Turnerj:
So now I've gone from "it builds in CI but doesn't build locally" to "it builds locally but doesn't build in CI"... 🤦‍♂️

RehanSaeed:
Perhaps we can add a CLI command to the tool to download the JSON-LD for us?

Turnerj:
It would have to be an _additional_ tool as we can't call the source generator one from CLI but yeah, we could have a thin schema download client. I still wouldn't want to call it in our own build process but it does simplify our updating process.

I'll mock something up as another commit to this PR.

Turnerj:
So I've...
- Moved the JSON-LD file to a new data folder in the root directory (seemed to make sense now we have two tools share it)
- Added a basic .NET 5 CLI application that will download the latest file from Schema.org and save it to this folder

Seems to work pretty good.

Turnerj:
I'm thinking in the future, using something [like this](https://github.com/marketplace/actions/github-push) with a scheduled GitHub Action to do this updating for us. That though is a whole separate thing we can handle another day.

RehanSaeed:
> I'm thinking in the future, using something [like this](https://github.com/marketplace/actions/github-push) with a scheduled GitHub Action to do this updating for us. That though is a whole separate thing we can handle another day.

We have to be careful with code that updates itself. We need to ensure there are no breaking changes, these normally fail the build/tests though. A better approach might be to have a GitHub action that raises a PR for us a bit like dependabot/renovate.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
```diff
@@ -3,12 +3,33 @@
...
     <TargetFramework>netstandard2.0</TargetFramework>
     <IsPackable>false</IsPackable>
+    <IsRoslynComponent>true</IsRoslynComponent>
   </PropertyGroup>
 
   <ItemGroup Label="Package References">
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.10.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(BuildingInsideVisualStudio)' == 'true'">
```
RehanSaeed, line 13: This looks pretty hacky. Worth adding a comment as to why we had to do this? Link to a GitHub issue maybe?

Tools/Schema.NET.Updater/Program.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    internal class Program
+    {
+        private const string SchemaJsonSourceUrl = "https://schema.org/version/latest/schemaorg-all-https.jsonld";
+        private const string SchemaJsonDestinationFilePath = "../../Data/schemaorg-all-https.jsonld";
+
+        private static async Task Main()
+        {
+            Console.WriteLine("Downloading '{0}'...", SchemaJsonSourceUrl);
+            var httpClient = new HttpClient();
```
RehanSaeed, line 19: Missing a `using` here.

RehanSaeed (commented):
Seems reasonable to me 👍🏼.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
Turnerj: Good idea about adding a comment - agreed the solution is pretty hacky. It might be a race to whether the Visual Studio/MSBuild handling "fixes" the issue or the .NET SDK does.

RehanSaeed:
It'd also then play nicely with the auto-generated release notes. There will be a manual step of clicking the publish button every so often though.

PR #302, opened 2021-06-21
Create Schema.NET.Pending project for pending types

### LINKED ISSUE #203: Add VirtualLocation type
### Describe the feature

It looks that the VirtualLocation type is added to schema.org recently. Can we get it added to the library?

### Schema Objects

Which [schema.org](https://schema.org) object is this about if any?

https://schema.org/VirtualLocation

Turnerj:
Will close #203 

The core changes:
- New "Schema.NET.Pending" project
- All the C# source files for Schema.NET have been moved to a "common" folder and referenced by it and the new project
- Minor changes to the source generator to pull in a setting about including pending types

I don't think I have this quite right yet (particularly around pulling in the setting) but is the general gist of the changes required to support pending types.

Once #301 is in, it will make debugging this easier and then hopefully we can get full pending types supported!


Turnerj:
A quick look suggests we might need another variation of our `Values` type with a different number of generic arguments. Will need to dig into this further.

Turnerj:
I've got it working! The builds will take about 2x longer now (likely also making Visual Studio a little more sluggish) but we get the functionality we want.

Source/Schema.NET.Pending/Schema.NET.Pending.csproj
```diff
@@ -0,0 +1,30 @@
...
+    <PackageReference Include="System.Memory" Version="4.5.4" />
+    <ProjectReference Include="../../Tools/Schema.NET.Tool/Schema.NET.Tool.csproj" OutputItemType="Analyzer" ReferenceOutputAssembly="false" />
+  </ItemGroup>
+
+  <ItemGroup Label="Package References (.NET Framework)" Condition="'$(TargetFramework)' == 'net461'">
+    <PackageReference Include="System.ValueTuple" Version="4.5.0" />
+  </ItemGroup>
+
+  <ItemGroup Label="Common Code">
+    <Compile Include="../Common/*.*" />
```
RehanSaeed, line 27: Took me a second to get my head around this, so there are common files added as solution level items and we're using compiler magic to include them. That seems fine, although I do wonder if it would be nicer to leave the files in Schema.NET and link to them from Schema.NET.Pending or even introduce a common assembly Schema.NET.Common.

RehanSaeed (approved):

Source/Schema.NET.Pending/Schema.NET.Pending.csproj
Turnerj: I thought linking them from "Schema.NET" to "Schema.NET.Pending" seemed to imply an inheritance/importance between them where they are actually equals. (I know that might sound weird but it made sense in my head) 

I thought about using a common project too. It would mean less weirdness adding new common files (and how that interacts with the `.sln` format) but we would run into problems if we actually want that common project to compile. The file `PropertyValueSpecification.Partial.cs` relies on the generated source. We could either duplicate this file manually or only reference it in our individual projects but I think we would be exchanging one weirdness for another.

Turnerj:
Once we've got this in, should we release a new version (and thus the first version of Schema.NET.Pending)?

Source/Schema.NET.Pending/Schema.NET.Pending.csproj
RehanSaeed: Makes sense. I haven't had a chance to play with this PR in VS, is there any weirdness with compiler errors/warnings this way? I assume not.

Turnerj: No errors or compiler warnings that I've seen. The only issue I've encountered is that solution folders don't act the same as folders in projects. If you right click -> add file to a solution folder, the file ends up in the root directory. If you right click -> add file to `Schema.NET` or `Schema.NET.Pending`, it would only add that file to that project.

RehanSaeed: Yeah, hate solution folders for that reason and the fact that the sln file is still indecipherable.

Turnerj:
Just want to make sure before I create a new release that the NuGet key you've configured for the repo has the "Create new packages" permission. I've had issues before of not doing that and then manually having to remove tags/releases and re-running it.

RehanSaeed:
Write permissions for GitHub packages are in the build:

image

I recently switched to using this least priviledge approach for all my repos and defaulted to read rights in settings:

image

Only downside is that autolabeller seems to have stopped working and I have no idea why:

https://github.com/release-drafter/release-drafter/issues/869

Turnerj:
Ahhhh, sorry, I meant permissions in NuGet.org itself. So when you have your API key in NuGet, you get a screen like this:

image

What I'm referring to is whether you have "Push new packages and package versions" or whether it is "Push new package versions". With our "Schema.NET.Pending" library, it will need that unless you want to upload the first package for it manually.

RehanSaeed:
Oh I see. I have one key per project and get notified when they expire, so we won't have that problem.

Turnerj:
Looks like we might have missed something. The Schema.NET.Pending package didn't push to NuGet and the one that pushed to GitHub Packages isn't correctly versioned... 🤔

Turnerj:
I think the `MinVerTagPrefix` is what caused the issue. It is `<MinVerTagPrefix>$(PackageId).</MinVerTagPrefix>` which won't work anymore with two packages with two different names (unless you want to do two tags per release - one `Schema.NET.<version>` and another `Schema.NET.Pending.<version>`).

RehanSaeed:
Fixed in https://github.com/RehanSaeed/Schema.NET/pull/303. Will delete the tag and re-release. Annoying but required.

Turnerj:
Cheers mate!

PR #311, opened 2021-08-10
Update .editorconfig from 2.1.0 to 3.0.0

RehanSaeed:
- See https://github.com/RehanSaeed/EditorConfig/releases/tag/3.0.0
- Set `<AnalysisMode>AllEnabledByDefault</AnalysisMode>`. This actually picked up a couple of small things that I've fixed. Added a bunch of warning suppressions too though. Also added `CLSCompliant(true)` so our F# folks can also play.
- In `SchemaService.cs` if we found a schema type that started with a number e.g. the `pending` type `3DModel` we prefixed it with an underscore. This is not CLSCompliant, so I changed the prefix to the word `Type`. Happy to pick something better.

### DIFF
--- a/Source/Common/ContextJsonConverter.cs
+++ b/Source/Common/ContextJsonConverter.cs
@@ -23,6 +23,11 @@ namespace Schema.NET
                 throw new ArgumentNullException(nameof(objectType));
             }
 
+            if (hasExistingValue && existingValue is null)
+            {
+                throw new ArgumentNullException(nameof(existingValue));
+            }
+
             if (serializer is null)
             {
                 throw new ArgumentNullException(nameof(serializer));
@@ -48,8 +53,10 @@ namespace Schema.NET
                 language = languageProperty?.Value?.ToString();
             }
 
+#pragma warning disable CA1062 // Validate arguments of public methods
             context.Name = name;
             context.Language = language;
+#pragma warning restore CA1062 // Validate arguments of public methods
             return context;
         }
 
--- a/Source/Common/EnumHelper.cs
+++ b/Source/Common/EnumHelper.cs
@@ -30,7 +30,9 @@ namespace Schema.NET
                 result = Enum.Parse(enumType, value);
                 return true;
             }
+#pragma warning disable CA1031 // Do not catch general exception types
             catch
+#pragma warning restore CA1031 // Do not catch general exception types
             {
                 result = null;
                 return false;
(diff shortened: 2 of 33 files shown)


Source/Common/EnumHelper.cs
```diff
@@ -30,7 +30,9 @@ internal static class EnumHelper
                 result = Enum.Parse(enumType, value);
                 return true;
             }
+#pragma warning disable CA1031 // Do not catch general exception types
             catch
+#pragma warning restore CA1031 // Do not catch general exception types
```
Turnerj, line 35: Is it worth moving this pragma (and the one below it about the expression body) to be for the whole file? Might be a little less noisy?

Source/Common/ContextJsonConverter.cs
```diff
@@ -48,8 +53,10 @@ public override JsonLdContext ReadJson(JsonReader reader, Type objectType, JsonL
                 language = languageProperty?.Value?.ToString();
             }
 
+#pragma warning disable CA1062 // Validate arguments of public methods
```
Turnerj, line 56: I'm confused by this warning - is this because those variables are the result of `ToString()` which is a public method?

Source/Common/Values{T1,T2,T3,T4,T5,T6,T7}.cs
```diff
@@ -503,7 +505,9 @@ public Values(IEnumerable<object?> items)
         /// <param name="list">The list of values.</param>
         /// <returns>The result of the conversion.</returns>
 #pragma warning disable CA2225 // Operator overloads have named alternates
+#pragma warning disable CA1002 // Do not expose generic lists
```
Turnerj, line 508: Is the warning wanting us to have `ICollection<T>` or `IList<T>` instead?

Also, similar to another comment, is it worth having this (and maybe the operator overload) pragma settings set for the whole file?

Tools/Schema.NET.Tool/Services/SchemaService.cs
```diff
@@ -220,7 +220,7 @@ private static void SetParentAndChildren(List<GeneratorSchemaClass> classes)
             HashSet<Uri> knownSchemaClasses,
             bool includePending)
         {
-            var className = StartsWithNumber.IsMatch(schemaClass.Label) ? $"_{schemaClass.Label}" : schemaClass.Label;
+            var className = StartsWithNumber.IsMatch(schemaClass.Label) ? $"Type{schemaClass.Label}" : schemaClass.Label;
```
Turnerj, line 223: I guess we just need to mention this in the release notes?

RehanSaeed: Only one pending type is affected.

Turnerj: Ahhhh, OK - not a big deal then!

RehanSaeed (commented):
Did a general cleanup of warning suppressions.

Turnerj (approved):
Looks good to me!

PR #322, opened 2021-08-21
Add TrySetValue and TryGetValue

Turnerj:
Introduces two new methods `TrySetValue` and `TryGetValue` which allow dynamic access to get and set values on a schema object. Both methods are generated via our source generator rather than doing something weird with reflection.

While `TrySetValue` is straight forward, `TryGetValue` requires a generic parameter to identify which piece of data you want. This is required to have a typed-value be the result of the call. If you specify a type that is invalid for the property, the method fails gracefully.

The methods are also case insensitive.

Open to any feedback or suggestions you have @RehanSaeed 🙂

### DIFF
--- a/Tests/Schema.NET.Test/ThingTest.cs
+++ b/Tests/Schema.NET.Test/ThingTest.cs
@@ -233,12 +233,59 @@ namespace Schema.NET.Test
                 Name = "TestName",
             };
 
+            Assert.True(thing.TryGetValue("Name", out var result));
+            var name = Assert.Single(result);
+            Assert.Equal("TestName", name);
+        }
+
+        [Fact]
+        public void TryGetValue_ValidProperty_Values()
+        {
+            var thing = new Thing
+            {
+                Identifier = new Uri("https://example.org/test-identifier"),
+            };
+
+            Assert.True(thing.TryGetValue("Identifier", out var result));
+            var identifier = Assert.Single(result);
+            Assert.Equal(new Uri("https://example.org/test-identifier"), identifier);
+        }
+
+        [Fact]
+        public void TryGetValue_InvalidProperty_InvalidName()
+        {
+            var thing = new Thing();
+
+            Assert.False(thing.TryGetValue("InvalidName", out _));
+        }
+
+        [Fact]
+        public void TryGetValue_CaseInsensitive()
+        {
+            var thing = new Thing
+            {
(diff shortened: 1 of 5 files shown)


Turnerj:
There is probably another method I could add too: `TryGetValue(string property, out IValues result)`
This being non-generic allows simple access to any type of value on the property.

Actually, I'm just going to add that now to this PR.

Tests/Schema.NET.Test/ThingTest.cs
```diff
@@ -199,6 +199,135 @@ public void ToStringWithNullAssignedProperty_ReturnsExpectedJsonLd()
...
+
+            Assert.False(thing.TrySetValue("InvalidName", new[] { "TestName" }));
+        }
+
+        [Fact]
+        public void TrySetValue_CaseInsensitive()
+        {
+            var thing = new Thing();
+
+            Assert.True(thing.TrySetValue("name", new[] { "TestName" }));
```
RehanSaeed, line 224: What made you go with case insensitivity?

RehanSaeed (commented):
Seems reasonable. Is there a specific reason you need this feature? I can't think of any obvious use cases.

Tests/Schema.NET.Test/ThingTest.cs
Turnerj: With the idea you don't know the properties at compile time, it seemed like the best idea for compatibility to support it.

Happy to have it be a setting you pass in instead - can default to case sensitive.

Turnerj:
> 
> 
> Seems reasonable. Is there a specific reason you need this feature? I can't think of any obvious use cases.

The primary motivation for me is that it makes it easier to create Schema objects at runtime - especially when the properties are provided from unreliable 3rd party data (for example, parsing Microdata from HTML into Schema objects).

Besides these methods, there are a couple of other changes I'd like to make for other methods of reading/writing/converting schema objects easier too. For example, moving some of the string-conversion logic in `ValuesJsonConverter` to a separate class that can be access directly (again, useful for things like converting Microdata).

Turnerj:
> ... there are a couple of other changes I'd like to make for other methods ...

I should clarify with this - I was originally going to include some of that in this PR but some of the changes would overlap with the changes for System.Text.Json support and would make merging a bit of a pain.

RehanSaeed (approved):

Issue #329, opened 2021-10-04
Regression in deserializing @language inside @context (from #63)


shravan2x:
### Describe the bug

In #63, the library added support for deserializing @context objects. But it appears that OP's original example doesn't deserialize any more. I've seen schema.org data in the wild that uses this format.

@RehanSaeed your comment https://github.com/RehanSaeed/Schema.NET/issues/63#issuecomment-500329801 mentioned the test case below but it wasn't included in the PR so did not trigger any failures. This test case does not pass any longer

```c#
private const string JsonWithLanguageInContext =
    "{\r\n" +
        "\"@context\": [ \"http://schema.org\", { \"@language\": \"it\" } ],\r\n" +
        "\"@type\": \"LocalBusiness\",\r\n" +
        "\"@id\": \"https://example.com/mario-rossi-srl\",\r\n" +
        "\"name\": \"Mario Rossi\"\r\n" +
    "}";

[Fact]
public void DeserializeObject_WithLanguageInContext_DeserializesSuccessfully()
{
    var localBusiness = JsonConvert.DeserializeObject<LocalBusiness>(JsonWithLanguageInContext);

    // Asserts....
}
```

The cause appears to be that [this line of code](https://github.com/RehanSaeed/Schema.NET/blob/main/Source/Common/ContextJsonConverter.cs#L47) tries to deserialize the array as an object, which fails. It could be fixed by adding a separate else-if for arrays.

### Steps to reproduce

1. Run the test case included above

### Expected behaviour

1. Model deserializes correctly.

RehanSaeed:
Thanks @shravan2x for raising this bug. We'd be willing to take a small PR for a fix.

Turnerj:
I was curious about whether that was a valid example and [while I found an example in the spec matching it](https://w3c.github.io/json-ld-syntax/#example-22-combining-external-and-local-contexts), I think there is a second issue with our context processing.

When we have the following:
```json
{
    "@context": { "name": "https://schema.org/" }
}
```

[The JSON-LD spec seems to imply it is a one-to-one mapping of the property "name" on the container.](https://w3c.github.io/json-ld-syntax/#example-11-term-expansion-from-context-definition)

I think JSON-LD contexts might actually be quite a bit more nuanced than we currently process.

shravan2x:
@Turnerj That's interesting, I didn't know. I can make a PR for array support but the second issue appears more complex than I have bandwidth for at the moment.

[...]

PR #331, opened 2021-10-09
Support deserializing @context arrays

shravan2x:
Per discussion in #329.


shravan2x:
The code is somewhat repetitive, but it seems we need more significant changes anyway (https://github.com/RehanSaeed/Schema.NET/issues/329#issuecomment-933364021). Happy to refactor if needed

PR #334, opened 2021-10-15
Add .NET 6 and C# 10 Support

RehanSaeed:
- Add embedded `README.md` to NuGet packages (https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packagereadmefile).
- Add `net6.0`.
- Remove `net461`.
- Remove unused code.
- Use `ArgumentNullException.ThrowIfNull(...)`.

Will update the `.editorconfig` in a later PR and go with file scoped namespaces.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,6 +33,14 @@ jobs:
         with:
           lfs: true
           fetch-depth: 0
+      - name: "Install .NET Core 3.1 SDK"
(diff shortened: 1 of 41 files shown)


Turnerj:
Source generators always have to target .NET Standard 2.0 unfortunately - it is a limitation baked into how they are invoked I believe. Might change when/if VS is no longer using .NET Framework.

Turnerj (approved):
LGTM!

Issue #389, opened 2022-02-09
Exception thrown The JSON value could not be converted to Schema.NET.IValues


dermotblairca:
### Describe the bug

The following exception is thrown when trying to parse the Product json in the following Walmart.com URL: https://www.walmart.com/ip/Air-Wick-Plug-in-Scented-Oil-Warmer-2-ct-White-color-Air-Freshener-Essential-Oils/14284640?athbdg=L1600
"System.Text.Json.JsonException: 'The JSON value could not be converted to Schema.NET.IValues. Path: $.name | LineNumber: 0 | BytePositionInLine: 29.'"
This validates fine on validtor.schema.org: https://validator.schema.org/#url=https%3A%2F%2Fwww.walmart.com%2Fip%2FAir-Wick-Plug-in-Scented-Oil-Warmer-2-ct-White-color-Air-Freshener-Essential-Oils%2F14284640%3Fathbdg%3DL1600

In my code, I am getting the Product json from the Script element and then calling the following:
`var product = SchemaSerializer.DeserializeObject<Schema.NET.Product>(json);`

I was also able to reproduce this by downloading the Schema.Net source and adding the following test to the ProductTest.cs file (as shown in Steps to reproduce):

### Steps to reproduce

Add the following to ProductTest.cs file and run the test:

`[Fact]
public void Deserializing_ProductJsonLd_ReturnsProduct2()
{
	try
	{
[...]

PR #390, opened 2022-02-09
Fix for null string values

RehanSaeed:
Fixes:

- https://github.com/RehanSaeed/Schema.NET/issues/389

### DIFF
--- a/Tests/Schema.NET.Test/ValuesJsonConverterTest.cs
+++ b/Tests/Schema.NET.Test/ValuesJsonConverterTest.cs
@@ -16,6 +16,14 @@ public class ValuesJsonConverterTest
     }
 
     [Fact]
+    public void WriteJson_ValuesNull_WritesNull()
+    {
+        var value = new Values<int?, string?>((string?)null);
+        var json = SerializeObject(value);
+        Assert.Equal("{}", json);
+    }
+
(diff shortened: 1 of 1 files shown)


Turnerj:
Looks like our issue is this block of code in S.T.J: https://source.dot.net/#System.Text.Json/System/Text/Json/Serialization/Metadata/JsonPropertyInfoOfT.cs,353

There are a couple of flags around there that we might be able to toggle to get this to behave correctly.

Turnerj:
I should mention - that the block of code I linked to is likely based on the master/main code in the Runtime, not necessarily what we are using. We don't have individual control of `HandleNullOnRead` currently, just a `HandleNull` property which does read and write.

I imagine (though haven't confirmed) is that if we enable `HandleNull`, it might mean that null/default values are going to be serialized again.

Turnerj:
Setting `HandleNull` does make our converter run however another internal part of S.T.J trips:
image

I'm thinking that S.T.J is doing this because a struct can't technically be null instead of setting the struct to its default value.

Turnerj:
Tried to mess around with a few settings but doesn't seem like there is anything we can do without trying something drastic like switching `Values` and `OneOrMany` to be reference types.

Issue #391, opened 2022-02-15
Relative paths are removed from Product


dermotblairca:
### Describe the bug

When a Product has a relative path, for example as the "image" field, this is not returned in the Product object from SchemaSerializer.DeserializeObject.
Maybe this was intentional but it does seem to validate ok in the Schema.org validator. For example: https://validator.schema.org/#url=https%3A%2F%2Fwww.thedandys.ie%2Fproduct%2Fsodium-hypochlorite-25ltrs%3Fgclid%3DCj0KCQiAu62QBhC7ARIsALXijXTUddi2aWyH-XjZKJHIuzUSLbH6HiVW883I7sSdYEtORXVeeALdfwkaAnALEALw_wcB
And the schema.org documentation shows relative paths in the examples too (see the example in the JSON-LD tab): https://schema.org/image

### Steps to reproduce

Run the following unit test in the ProductTest.cs file:
```
namespace Schema.NET.Test.Examples;

using System;
using Xunit;

// https://developers.google.com/search/docs/data-types/products
public class ProductTest
{
    [Fact]
    public void Deserializing_ProductJsonLd_ReturnsProduct2()
    {
        var productJson = "{\"@context\":\"http://www.schema.org\",\"@type\":\"product\",\"name\":\"SodiumHypochlorite/RedLabel25Ltrs\",\"image\":\"/site/uploads/sys_products/red-label-hypo.jpg\",\"offers\":{\"@type\":\"Offer\",\"price\":\"22.00\",\"availability\":\"http://schema.org/InStock\",\"priceCurrency\":\"EUR\"}}";
        var productObject = SchemaSerializer.DeserializeObject<Product>(productJson);
        var productBackToString = productObject?.ToString();
        var hasImagePath = productBackToString?.IndexOf("/site/uploads/sys_products/red-label-hypo.jpg", StringComparison.OrdinalIgnoreCase) > 0;
        Assert.True(hasImagePath);
    }
}
```

### Expected behaviour

The image should be included in the Product object.

### Schema objects

https://schema.org/Product (but I think this would affect any class derived from Thing)

Turnerj:
Thanks for raising this - we've seen this come up already in #343. It is a bit of a weird one to solve.

dermotblairca:
That's interesting, thanks for the link to the discussion @Turnerj

RehanSaeed:
[...]

PR #401, opened 2022-02-20
Use UriKind.RelativeOrAbsolute to allow deserializing of relative URIs

dermotblairca:
Fixing issue where relative URIs in JSON are ignored and not included in the deserialized schema.org objects, as discussed in #391

### DIFF
--- a/Source/Common/ValuesJsonConverter.cs
+++ b/Source/Common/ValuesJsonConverter.cs
@@ -379,7 +379,7 @@ public class ValuesJsonConverter : JsonConverter<IValues>
             }
             else if (targetType == typeof(Uri))
             {
-                success = Uri.TryCreate(valueString, UriKind.Absolute, out var localResult);
+                success = Uri.TryCreate(valueString, UriKind.RelativeOrAbsolute, out var localResult);
                 result = localResult;
             }
             else if (targetType == typeof(Guid))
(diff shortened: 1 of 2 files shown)


RehanSaeed:
Can we also add some tests?

dermotblairca:
Hi @RehanSaeed I have added a new test to check the relative URI scenario, and also updated an existing test to include a relative URI when deserializing an object of type Thing. It looks like the absolute URI scenarios are already sufficiently tested. However if you would like more test cases, let me know and I can add them. Thanks.

RehanSaeed:
Thanks @dermotblairca. The `ReadJson_Values_SingleValue_ThingInterface` test seems to be failing though.

dermotblairca:
@RehanSaeed I can't actually reproduce that failed test. When I do a clean rebuild on my fork, all of the tests run fine including that one. For example:
image
Could it be environment-specific maybe?

dermotblairca:
Hopefully my last change has fixed the tests. I think when ToString() was implicitly called on the URI of actual/second argument on the Assert call, it seems to prepend it with "file://". I cannot reproduce this though so I'm guessing it only happens on some environments.

RehanSaeed:
Thanks!

dermotblairca:
No problem!

PR #465, opened 2022-07-27
Upgrade to .NET 7 and C# 11

RehanSaeed:
- Add .NET 7 target.
- Drop .NET 5 and 3.1 target.
- Update build scripts to install .NET 7 and 6 and stop installing 5 and 3.1.
- Uses C# 11 features resulting in cleaner code. Raw string literals in particular.
- Added `#nullable enable` to generated code which is required to enable nullable reference types. I thought we'd already done it but turns out not!
- Generate simpler code with better performance.
- Switch `dotnet test --logger` to use `LogFilePrefix`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,7 +38,7 @@ jobs:
         with:
           dotnet-version: |
             3.1.x
-            5.0.x
+            6.0.x
           global-json-file: "./global.json"
       - name: "Dotnet Tool Restore"
         run: dotnet tool restore
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,6 @@ jobs:
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: |
-            3.1.x
             6.0.x
           global-json-file: "./global.json"
       - name: "Dotnet Tool Restore"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,6 +73,7 @@ jobs:
           path: "./Artefacts"
       - name: "Publish Test Summary"
         uses: test-summary/action@v2
+        if: always()
         with:
           paths: "./Artefacts/*/*.xml"
 
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,9 +25,9 @@
   </PropertyGroup>
 
   <ItemGroup Label="Package References">
-    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="All" Version="17.4.27" />
-    <PackageReference Include="MinVer" PrivateAssets="All" Version="4.2.0" />
-    <PackageReference Include="StyleCop.Analyzers" PrivateAssets="All" Version="1.2.0-beta.435" />
+    <GlobalPackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.4.27" />
+    <GlobalPackageReference Include="MinVer" Version="4.2.0" />
+    <GlobalPackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" />
   </ItemGroup>
 
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,9 +25,9 @@
   </PropertyGroup>
 
   <ItemGroup Label="Package References">
-    <GlobalPackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" Version="17.4.27" />
-    <GlobalPackageReference Include="MinVer" Version="4.2.0" />
-    <GlobalPackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.435" />
+    <PackageReference Include="Microsoft.VisualStudio.Threading.Analyzers" PrivateAssets="All" Version="17.4.27" />
+    <PackageReference Include="MinVer" PrivateAssets="All" Version="4.2.0" />
+    <PackageReference Include="StyleCop.Analyzers" PrivateAssets="All" Version="1.2.0-beta.435" />
   </ItemGroup>
 
 </Project>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup Label="Build">
-    <LangVersion>latest</LangVersion>
+    <LangVersion>preview</LangVersion>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
     <AnalysisLevel>latest</AnalysisLevel>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup Label="Build">
-    <LangVersion>preview</LangVersion>
+    <LangVersion>latest</LangVersion>
     <EnforceCodeStyleInBuild>true</EnforceCodeStyleInBuild>
     <AnalysisMode>AllEnabledByDefault</AnalysisMode>
     <AnalysisLevel>latest</AnalysisLevel>
--- a/Source/Common/PropertyValueSpecification.Partial.cs
+++ b/Source/Common/PropertyValueSpecification.Partial.cs
@@ -13,7 +13,7 @@ public partial class PropertyValueSpecification
     private const string NamePropertyName = "name=";
     private const string PatternPropertyName = "pattern=";
     private const string RequiredPropertyName = "required";
-    private const char Space = ' ';
+    private const char SpaceDelimeter = ' ';
 
     /// <summary>
     /// Returns a <see cref="string" /> that represents the short hand representation of this instance.
@@ -34,39 +34,39 @@ public partial class PropertyValueSpecification
 
         if (this.ValueMinLength.First() is double minLength)
         {
-            AppendSpace(stringBuilder);
+            AppendSpaceDelimeter(stringBuilder);
             stringBuilder.Append(MinLengthPropertyName);
             stringBuilder.Append(minLength);
         }
 
         if (this.ValueName.First() is string name)
         {
-            AppendSpace(stringBuilder);
+            AppendSpaceDelimeter(stringBuilder);
             stringBuilder.Append(NamePropertyName);
             stringBuilder.Append(name);
         }
 
         if (this.ValuePattern.First() is string pattern)
         {
-            AppendSpace(stringBuilder);
+            AppendSpaceDelimeter(stringBuilder);
             stringBuilder.Append(PatternPropertyName);
             stringBuilder.Append(pattern);
         }
 
         if (this.ValueRequired.First() is true)
         {
-            AppendSpace(stringBuilder);
+            AppendSpaceDelimeter(stringBuilder);
             stringBuilder.Append(RequiredPropertyName);
         }
 
         return stringBuilder.ToString();
     }
 
-    private static void AppendSpace(StringBuilder stringBuilder)
+    private static void AppendSpaceDelimeter(StringBuilder stringBuilder)
     {
         if (stringBuilder.Length > 0)
         {
-            stringBuilder.Append(Space);
+            stringBuilder.Append(SpaceDelimeter);
         }
     }
 }
--- a/Source/Common/PropertyValueSpecification.Partial.cs
+++ b/Source/Common/PropertyValueSpecification.Partial.cs
@@ -1,6 +1,5 @@
 namespace Schema.NET;
 
-using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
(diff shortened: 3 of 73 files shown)


Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup Label="Build">
-    <LangVersion>latest</LangVersion>
+    <LangVersion>preview</LangVersion>
```
RehanSaeed, line 4: Change this back to latest when .NET 7 hits production.

global.json
```diff
@@ -2,6 +2,6 @@
   "sdk": {
     "allowPrerelease": true,
     "rollForward": "latestMajor",
-    "version": "6.0.300"
+    "version": "7.0.100-preview.6.22352.1"
```
RehanSaeed, line 5: Change this when .NET 7 hits production.

Source/Common/SchemaEnumJsonConverter{T}.cs
```diff
@@ -41,11 +41,21 @@ public SchemaEnumJsonConverter()
...
+#else
         if (typeToConvert is null)
         {
             throw new ArgumentNullException(nameof(typeToConvert));
         }
 
+        if (options is null)
+        {
+            throw new ArgumentNullException(nameof(options));
+        }
```
Turnerj, line 56: Does it make sense to assert this if we don't actually use `options`?

Source/Common/PropertyValueSpecification.Partial.cs
```diff
@@ -9,6 +8,13 @@ namespace Schema.NET;
 /// </summary>
 public partial class PropertyValueSpecification
 {
+    private const string MaxLengthPropertyName = "maxlength=";
+    private const string MinLengthPropertyName = "minlength=";
+    private const string NamePropertyName = "name=";
+    private const string PatternPropertyName = "pattern=";
+    private const string RequiredPropertyName = "required";
+    private const char Space = ' ';
```
Turnerj, line 16: The others above seem reasonable but bringing the space out seems a little strange.

Tests/Schema.NET.Test/StringExtensions.cs
```diff
@@ -1,13 +1,36 @@
...
 namespace Schema.NET.Test;
 
+#if NET472 || NET461
 using System;
-
-#pragma warning disable IDE0060 // Remove unused parameter
-#pragma warning disable CA1801 // Remove unused parameter
+#endif
+using System.IO;
+using Newtonsoft.Json;
```
Turnerj, line 7: Not sure this is right - shouldn't this be `System.Text.Json`?

Turnerj: I know this is just the test project but probably makes sense to use the same library throughout the project.

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaClass.cs
```diff
@@ -41,7 +41,7 @@ public GeneratorSchemaClass(string layer, Uri id, string name, string descriptio
 
     public bool IsCombined { get; }
 
-    public IEnumerable<GeneratorSchemaProperty> DeclaredProperties
+    public ICollection<GeneratorSchemaProperty> DeclaredProperties
```
Turnerj, line 44: Maybe use `IReadOnlyCollection` (as we aren't intending someone to add to this)

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaClass.cs
```diff
@@ -50,7 +50,7 @@ public IEnumerable<GeneratorSchemaProperty> DeclaredProperties
             var declaredProperties = this.Properties.Where(classProp => !ancestorProps.Contains(classProp.Name))
                 .OrderBy(x => x.Order);
 
-            return declaredProperties;
+            return declaredProperties.ToList();
```
Turnerj, line 53: Rather than `ToList`, let's use `ToArray`. It implements `IReadOnlyCollection` and avoids the (admittedly small) allocation overhead of `List`.

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaProperty.cs
```diff
@@ -28,7 +28,7 @@ public GeneratorSchemaProperty(GeneratorSchemaClass @class, string jsonName, str
 
     public ICollection<GeneratorSchemaPropertyType> Types { get; } = new List<GeneratorSchemaPropertyType>();
 
-    public IEnumerable<string> CSharpTypes => this.Types.SelectMany(x => x.CSharpTypeStrings);
+    public ICollection<string> CSharpTypes => this.Types.SelectMany(x => x.CSharpTypeStrings).ToList();
```
Turnerj, line 31: Ditto of my previous comment about `IReadOnlyCollection`/`ToArray`

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
```diff
@@ -95,174 +141,271 @@ private static string RenderClass(GeneratorSchemaClass schemaClass)
...
+            }
+            """,
+            indent);
     }
 
-    [SuppressMessage(
-        "StyleCop.CSharp.LayoutRules",
-        "SA1513:Closing brace should be followed by blank line",
-        Justification = "Interpolated string")]
+    private static string RenderClassTryGetVaue1(GeneratorSchemaProperty[] allProperties, int indent)
```
Turnerj, line 231: Typo in method name - thinking maybe it would be better called `RenderClassTryGetValue_OneOrMany`.

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
```diff
@@ -95,174 +141,271 @@ private static string RenderClass(GeneratorSchemaClass schemaClass)
...
+                    success = base.TryGetValue(property, out result);
+                }
+
+                return success;
+            }
+            """,
+            indent);
+    }
+
+    private static string RenderClassTryGetVaue2(GeneratorSchemaProperty[] allProperties, int indent)
```
Turnerj, line 302: Typo in method name - thinking maybe it would be better called `RenderClassTryGetValue_IValues`.

Tools/Schema.NET.Tool/SourceUtility.cs
```diff
@@ -39,37 +46,31 @@ public static string RenderItems<T>(IEnumerable<T> items, Func<T, string> action
...
-#endif
+            stringBuilder.Append(Space, indent);
 
-        var stringBuilder = new StringBuilder();
-        var i = 0;
-        foreach (var item in items)
-        {
-            stringBuilder.Append(action(item, i));
-            i++;
+#pragma warning disable IDE0072 // Add missing cases
```
Turnerj, line 57: I'm surprised we need to disable this as you've got the discard/catch-all rule - do you know what's up with that?

Turnerj (changes requested):

Source/Common/PropertyValueSpecification.Partial.cs
RehanSaeed: Is that because its a `char`? Do you not normally make those constants?

Source/Common/SchemaEnumJsonConverter{T}.cs
RehanSaeed: Good catch. Got overzealous here. Removed.

Tests/Schema.NET.Test/StringExtensions.cs
RehanSaeed: Yes, I meant to do this but hadn't yet worked out how to implement the function using System.Text.Json. Leaving open.

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaClass.cs
RehanSaeed: Done

Tools/Schema.NET.Tool/GeneratorModels/GeneratorSchemaProperty.cs
RehanSaeed: Done

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
RehanSaeed: Done

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
RehanSaeed: Done

Tools/Schema.NET.Tool/SourceUtility.cs
RehanSaeed: I think this is a VS bug:

image

I've noticed a lot of times in the past where I've had to ignore warnings only to find that the warning no longer occurs. This usually happens when a new .NET version is around the corner.

Tests/Schema.NET.Test/StringExtensions.cs
RehanSaeed: Ok figured it out.

Source/Common/PropertyValueSpecification.Partial.cs
Turnerj: Yeah - like I think where constants shine are when:
- You have multiple instances of it (this doesn't always matter but is usually a good reason)
- It isn't immediately obvious _why_ the value is what is it (this is typically the main reason)
- The value is significant (eg. a long value that doesn't fit inline very well to the code)

Admittedly I could say _all_ of the constants here don't meet these requirements but also it can be an amount of personal preference too.

Like, nothing seems too "magic" about the values to need to be brought up to constants but nothing seems too bad about doing it.

Where my thoughts get extra complicated on it is _where_ the constants are. For example, maybe these as constants would make more sense inline right where they're used rather than on the class itself.

To be clear though, if you prefer it this way, happy to go that direction even if it is atypical to what I might do. 🙂

Turnerj: OK, it turns out that I just made a `char` a constant in some code I'm working on. In this case, my constants were like:

```csharp
private const char PathEndSpecialCharacter = '$';
private const char WildcardSpecialCharacter = '*';
```

They described the purpose or perhaps the meaning of the value rather than the value.

RehanSaeed: Renamed `Space` to `SpaceDelimeter`.

Also updated to the latest .NET 7 preview release.

Turnerj (approved):

RehanSaeed:
@Turnerj Updated to RTM. Was getting scary looking warnings about 3.1 not being supported and being untested, so I also removed that. Also, upgraded net461 to net462 for the same reason. Finally, made these two additional updates I discovered:

- Use GlobalPackageReference.
- Switch dotnet test --logger to use LogFilePrefix.

.github/workflows/build.yml
```diff
@@ -37,8 +37,7 @@ jobs:
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: |
-            3.1.x
-            5.0.x
+            6.0.x
```
Turnerj, line 40: Should this include `7.0.x` too?

RehanSaeed: The line below `global-json-file: "./global.json"` installs 7, so no need.

Tools/Schema.NET.Tool/SchemaSourceGenerator.cs
```diff
@@ -85,6 +80,57 @@ private static string RenderClass(GeneratorSchemaClass schemaClass)
...
+
+            {{SourceUtility.RenderItems(allProperties, RenderClassProperty, 4, SourceDelimeter.NewLineSpace)}}
+
+            {{RenderClassTrySetValue(allProperties, 4)}}
+
+            {{RenderClassTryGetVaueOneOrMany(allProperties, 4)}}
+
+            {{RenderClassTryGetVaueIValues(allProperties, 4)}}
+
+            {{RenderClassEquals(schemaClass, allProperties, 4)}}
```
Turnerj, line 119: Is it just me or does the indenting look wrong for these? Think it needs to be one further in.

RehanSaeed: Yeah, I think we could refactor this to make it read better. The fact is, these functions indent each line of code they generate by the number of characters supplied. Perhaps it could be better if they did not indent the first line.

RehanSaeed: Ok if we leave this for now? I'm currently renovating my house and have little time to contribute.

RehanSaeed: I'm currently renovating my house and am not going to have time to resolve this issue. @Turnerj Is it ok if we merge this and release what we have for now?

Turnerj: Yeah, all good mate 🙂

PR #548, opened 2022-12-23
Support JSON comments

### LINKED ISSUE #416: Cannot deserialize json that contains comment
### Describe the bug

If the json has a comment in it, SchemaSerializer.DeserializeObject throws the exception `'/' is an invalid start of a property name. Expected a '"'`
It looks like this can be resolved by adding the following to the JsonSerializerOptions: `ReadCommentHandling = JsonCommentHandling.Skip`

### Steps to reproduce

Try deserialize json with a comment in it. This can be demonstrated by the unit test below:
```
[Fact]
    public void Deserializing_ProductJsonLd_HasComment()
    {
        var jsonWithComment =
            "{" +
                "\"@context\":\"https://schema.org\"," +
                "\"@type\":\"Product\"," +
                "\"name\":\"Executive Anvil\", // some comment" +
                "\"description\":\"Sleeker than ACME's Classic Anvil, the Executive Anvil is perfect for the business traveller looking for something to drop from a height.\"," +
                "\"image\":\"https://www.example.com/anvil_executive.jpg\"," +
            "}";

        var exceptionMessage = string.Empty;
        try
        {
            SchemaSerializer.DeserializeObject<Product>(jsonWithComment);
        }
        catch (System.Text.Json.JsonException ex)
        {
            exceptionMessage = ex.Message;
        }

        Assert.Equal(string.Empty, exceptionMessage);
    }
```

### Expected behaviour

Deserializes the json without throwing an exception

### Schema objects

https://schema.org/Product (but affecting all objects)

Turnerj:
Closes #416


RehanSaeed (dismissed):
Looks good

PR #590, opened 2023-05-17
Upgrade to Schema.org v16

Turnerj:
Upgrades to Schema.org v16 (from v12)

This also includes a fix of sorts for the updater tool where if you ran it from Visual Studio, it wouldn't be the right directory. It now passes the project directory via MSBuild which is then read from the assembly attributes.

Additionally needed to add a `Values{T1,T2,T3,T4,T5}` as the pending types now require it.


Tools/Schema.NET.Updater/Program.cs
```diff
@@ -22,9 +24,23 @@ private static async Task Main()
...
 
-        using var fileStream = File.Open(SchemaJsonDestinationFilePath, FileMode.Create);
+        var fullSavePath = Path.Combine(GetProjectDirectory() ?? string.Empty, SchemaJsonDestinationFilePath);
+        using var fileStream = File.Open(fullSavePath, FileMode.Create);
         await stream.CopyToAsync(fileStream).ConfigureAwait(true);
 
         Console.WriteLine(Resources.UpdateComplete);
     }
+
+    private static string? GetProjectDirectory()
```
RehanSaeed, line 34: Tried with Linq, don't think its much more readable.

```c#
private static string? GetProjectDirectory() =>
    Attribute
        .GetCustomAttributes(typeof(Program).Assembly)
        .OfType<AssemblyMetadataAttribute>()
        .Where(x => x.Key == ProjectDirectoryAttributeKey)
        .Select(x => x.Value)
        .FirstOrDefault();
```

RehanSaeed (approved):

Tools/Schema.NET.Updater/Program.cs
Turnerj: Yeah, was going to do similar but I think the plain foreach comes out slightly ahead in readability.

Turnerj:
Looks like two build issues remain:

> Ambiguous user defined conversions 'Values<IEntryPoint, Uri, Uri>.implicit operator Values<IEntryPoint, Uri, Uri>(Uri)' and 'Values<IEntryPoint, Uri, Uri>.implicit operator Values<IEntryPoint, Uri, Uri>(Uri)' when converting from 'Uri' to 'Values<IEntryPoint, Uri, Uri>'

This looks like it is coming from the `Target` property on `Action`. Guessing maybe a mistake in the schema.org JSON-LD file. We probably can tweak our tool to be distinct per type to avoid this bug.

> 'Course' does not contain a definition for 'Provider'

Likely the property has been removed from schema.org but I haven't gone to check yet.

RehanSaeed:
Its surprising how often they have bugs in their JSON-LD. You'd think they'd add a test or two.

Tests/Schema.NET.Test/Examples/CourseTest.cs
```diff
@@ -10,7 +10,7 @@ public class CourseTest
     {
         Name = "Introduction to Computer Science and Programming", // Required
         Description = "Introductory CS course laying out the basics.", // Required
-        Provider = new Organization() // Recommended
+        SourceOrganization = new Organization() // Note: This is typically set via "Provider" (a pending property) and is required
```
Turnerj, line 13: The `Provider` property is marked as pending now so isn't available in our tests (which use the `Schema.NET`, not `Schema.NET.Pending`

Tools/Schema.NET.Tool/CustomOverrides/AddTextTypeToActionTarget.cs
```diff
@@ -1,38 +0,0 @@
...
-        ArgumentNullException.ThrowIfNull(c);
-#else
-        if (c is null)
-        {
-            throw new ArgumentNullException(nameof(c));
-        }
-#endif
-
-        var property = c.Properties.First(x => string.Equals(x.Name, "target", StringComparison.OrdinalIgnoreCase));
-        property.Types.Add(new GeneratorSchemaPropertyType("URL", "Uri"));
```
Turnerj, line 36: This was where we got the duplicate `Uri` types from - guessing this means that Schema.org has included URI as an official type for the target property.

RehanSaeed (approved):

PR #607, opened 2023-06-08
Split target of dotnet library versions

### LINKED ISSUE #598: Version 12.0.0 has a dependency on .NET 7 in the .NET 6 build
### Describe the bug

We have to stay on .NET 6 for now, because of some Azure Durable Entities issues, and I'm unable to upgrade to the latest Schema.NET because it's pulling in System.Text.Json 7.0.0, from the .NET 6.0 target.

### Steps to reproduce

image

### Expected behaviour

.NET 6.0 target shouldn't pull in any .NET 7 libraries

### Schema objects

_No response_

Turnerj:
This will make only .NET 7 pull in v7 of core .NET libraries. Fixes #598 

The downside to this is that Renovate and Dependabot will likely struggle updating System.Text.Json now so we will need to be extra careful.

### DIFF
--- a/Source/Schema.NET.Pending/Schema.NET.Pending.csproj
+++ b/Source/Schema.NET.Pending/Schema.NET.Pending.csproj
@@ -13,10 +13,14 @@
     <PackageTags>Schema;.NET;Schema.org;Schema.NET;Structured Data;Google Structured Data</PackageTags>
   </PropertyGroup>
 
-  <ItemGroup Label="Package References">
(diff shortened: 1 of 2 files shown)


kirk-marple:
@RehanSaeed Would there be a chance to get this merged in?  We've been holding off upgrading to latest package for months because of this, and would really appreciate it, thanks!

RehanSaeed (approved):

RehanSaeed:
Strange that we have to do this. Approved.

Issue #675, opened 2023-12-13
A null or whitespace string passed into Values<T1, T2*> causes HasValueX to return true


SmithPlatts:
### Describe the bug

When the `Values(params object[] items)` or `Values(IEnumerable<object?> items)` constructors are used, and:

- when the target `Tn` is a string and the value is either `null` or a whitespace string; or
- when the target `Tn` is a nullable type; then

The constructor logic will evaluate `HasValueX` before assigning to `ValueX`, where `ValueX` returns an empty collection post-`OneOrMany<Tn>` cast.

### Steps to reproduce

A simple reproducible sample can be achieved using https://schema.org/Reservation:

``` csharp
string json = """
{
  "@context": "https://schema.org",
  "@type": "Reservation",
  "reservationId": "1",
  "totalPrice": ""
}
""";

Schema.NET.Reservation reservation = Schema.NET.SchemaSerializer.DeserializeObject<Schema.NET.Reservation>(json)!;
_ = reservation.TotalPrice.Value3.Count == 0 ? "Correct, totalPrice should be an empty collection" : throw new InvalidDataException("totalPrice should be an empty collection");
_ = !reservation.TotalPrice.HasValue3 ? "Correct, totalPrice should not have any values" : throw new InvalidDataException("totalPrice should return false for HasValue3");
```

1. Deserialise simple `Reservation` where `totalPrice` is an empty string.
2. Test that `Reservation.TotalPrice` has no values for values property 3 (`string`).
3. Test that `Reservation.TotalPrice` returns `false` for `HasValue3` (optionally the same test for `HasValue`).
4. Observe that `InvalidDataException` is thrown with the message `totalPrice should return false for HasValue3`.

### Expected behaviour

`Reservation.TotalPrice.HasValue3` should return `false`, not `true`.

### Schema objects

- All/any

PR #676, opened 2023-12-14
Fix HasValueX incorrectly returning true

SmithPlatts:
# Overview

As per issue #675, when a whitespace string is passed to the `Values(params object[] items)` or `Values(IEnumerable<object?> items)` constructors of any of the `Values<>` classes, the `HasValueX` property will return `true`, where the corresponding `ValueX` will be empty.

# Primary changes

- Added tests to handle the known failure scenario.
- Moved `this.HasValueX` evaluation and assignment to after `this.ValueX` assignment.

# Ancillary changes

- Updated `.editorconfig` to redefine new Roslyn analysers, ensuring that newer warnings don't fail the build.
- Added a helper method to `Schema.NET.Test`, `AssertEx.Empty<T>(OneOrMany<T> collection)`, which addresses the instances where the compiler was implicitly using `string` for `Assert.Empty(OneOrMany<string>)`, instead of treating the input as an enumerable.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -84,6 +84,10 @@ spelling_checkable_types =            # NEW
 # Default Severity for all .NET Code Style rules below
 dotnet_analyzer_diagnostic.severity = warning
 
+# VSSpell001: Spell Check
+dotnet_diagnostic.VSSpell001.severity = suggestion
+dotnet_diagnostic.VSSpell002.severity = suggestion
+
 ##########################################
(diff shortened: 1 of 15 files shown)


Turnerj:
Hey @SmithPlatts - thanks for this! I'll try and review it soon as what you described in this and corresponding issue make sense.

Turnerj (approved):
Looks good to me!

PR #696, opened 2025-01-12
Added .NET 8 target

multiarc:
1. Added .NET 8 Support
   - Aligned package versions with respective targets (.NET 6, netstandard2.0, .NET 8.0)
   - Removed .NET 7.0 target and outdated references
   - Removed direct targets on the library for .NET Framework and only left netstandard2.0 because of dependencies, netstandard2.0 reference required anyways and there is no point to split targets on the library itself.

2. Removed StyleCop.Analyzers in favor of Microsoft.CodeAnalysis.NetAnalyzers (debatable, but feels appropriate in this case); the reason being that StyleCop analyzers (even custom non-deprecated builds) are not always aware about newest C# features for .NET 8 target.
3. Fixed nullable references issues and other minor issues
    - Because struct is used as a base wrapper for everything it effectively made code incompatible with previous implementation when underlying value is any ValueType, because all input types became nullable and all output types became non-nullable. I'm not sure if making this breaking change is proper direction.

5. Moved packages for unit tests out of .props file
6. Upgraded xUnit
   - Fixed xUnit warnings

7. Upgraded the rest of packages in unit tests
8. Added packages packages.lock.json for repeatable restore


multiarc:
It appears that build runners do not support .NET 8 SDK.

azure-pipelines.yml
```diff
@@ -43,21 +43,16 @@ stages:
...
             displayName: "Install .NET Core 6.0 SDK"
             inputs:
               packageType: "sdk"
               version: 6.0.x
           - task: UseDotNet@2
-            displayName: "Install .NET Core SDK"
+            displayName: "Install .NET Core 8.0 SDK"
             inputs:
               packageType: "sdk"
-              useGlobalJson: true
```
Turnerj, line 60: What made you want to switch from the global.json being used to specifying the .NET 8 SDK manually? Just consistency with the others?

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
```diff
@@ -1,62 +1,42 @@
...
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGSystem_Text_Encodings_Web)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGSystem_Buffers)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGSystem_Memory)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGSystem_Numerics_Vectors)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGSystem_Runtime_CompilerServices_Unsafe)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGMicrosoft_Bcl_AsyncInterfaces)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-      <TargetPathWithTargetPlatformMoniker Include="$(PKGSystem_Threading_Tasks_Extensions)\lib\netstandard2.0\*.dll" IncludeRuntimeDependency="false" />
-    </ItemGroup>
-  </Target>
+<!--  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">-->
```
Turnerj: This whole section was to work around transitive dependency issues with source generators (basically if your dependencies have dependencies, you need to specify them all for source generators). Did you not have issues with this with the code compiling?

I guess my only thought is the lock files may help resolve this but I haven't seen anyone mention that on the discussions on GitHub about source generators before.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
```diff
@@ -1,62 +1,42 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;netstandard2.0</TargetFrameworks>
```
Turnerj, line 4: Source generators can only target `netstandard2.0` unfortunately.

Tests/Schema.NET.Test/Schema.NET.Test.csproj
```diff
@@ -1,13 +1,27 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup Label="Build">
-    <TargetFrameworks>net7.0;net6.0;net472</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;net462;net472;net48</TargetFrameworks>
```
Turnerj: Not sure how necessary it is to test 3 different versions of .NET Framework. I'd personally be happy to just test one .NET Framework version (likely 4.8).

.github/workflows/build.yml
```diff
@@ -33,12 +33,14 @@ jobs:
...
+        uses: actions/setup-dotnet@v4.2.0
         with:
-          dotnet-version: |
-            6.0.x
-          global-json-file: "./global.json"
+          dotnet-version: 6.0.x
+      - name: "Install .NET 8.0 SDK"
+        uses: actions/setup-dotnet@v4.2.0
+        with:
+          dotnet-version: 8.0.x
```
Turnerj: You can specify multiple dotnet versions in a single setup-dotnet call: https://github.com/actions/setup-dotnet?tab=readme-ov-file#usage

Turnerj (changes requested):
Hey - appreciate the effort you went to with this PR. I've added some comments.

azure-pipelines.yml
multiarc: Ideally, you would use latest available patch when building project as opposed to whatever is laying around in your global.json, unless there is a compatibility issue. This was the reason of removal

Tests/Schema.NET.Test/Schema.NET.Test.csproj
multiarc: I don't see a harm of listing major relevant versions in here. Benchmarks is another story and probably it worth removing. But I agree, generally speaking it shouldn't matter.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
multiarc: This one is interesting. Purely speaking you can use any target for source generator: after all this is just Roslyn, you have all targets available. But! It only works in console (dotnet command line) and in Rider IDE. Visual Studio and MSBuild cannot do that properly and spawns all sorts of issues (missing references, not able to compile, etc.) 
image
As opposed to command line (view from Rider):
image

multiarc: I can even remove netstandard2.0 completely from the targets and it continues to function no problem.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
multiarc: Do you have repro steps for this? I will check and add some tests as well to cover this case.

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
multiarc: I will look into it.

multiarc: I split Tool project into separate projects to allow proper package versioning for different targets when built from console (and published as a package) - as you cannot properly reference different versions of System.Text.Json and others from within single target. While allowing usage from Visual Studio and falling back to netstandard2.0 only.

Tests/Schema.NET.Test/Schema.NET.Test.csproj
multiarc: Please let me know how to proceed here. Keep all net versions or leave net48 only?

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
multiarc: I think I've got this reproduced in Visual Studio and fixed. I tested Rider, Visual Studio and console builds

Tools/Schema.NET.Tool/Schema.NET.Tool.csproj
multiarc: I fixed build for Visual Studio, Rider and CLI

Tests/Schema.NET.Test/Schema.NET.Test.csproj
multiarc: Removed redundant versions of older .NET framework

